PR #1, opened 2015-07-04
Add skeletal setup.py

jakewins:


setup.py
```diff
@@ -0,0 +1,30 @@
+
```
nigelsmall: Could you add the standard header containing the license?

setup.py
```diff
@@ -0,0 +1,30 @@
...
+    "long_description": "", 
+    "author": "Neo Technology",
+    "classifiers": [
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Operating System :: OS Independent",
+        "Topic :: Database",
+        "Topic :: Software Development",
+    ],
+    "zip_safe": False,
```
nigelsmall: No need to have this. It's zip_safe :-)

setup.py
```diff
@@ -0,0 +1,30 @@
...
+    "version": "1.0.0",
+    "description": "Python driver for Neo4j",
+    "long_description": "", 
+    "author": "Neo Technology",
+    "classifiers": [
+        "Intended Audience :: Developers",
+        "License :: OSI Approved :: Apache Software License",
+        "Operating System :: OS Independent",
+        "Topic :: Database",
+        "Topic :: Software Development",
```
nigelsmall: Should add language trove classifiers (https://pypi.python.org/pypi?%3Aaction=list_classifiers)

Aim to support Python 2.6, 2.7, 3.3, 3.4 and 3.5 plus Jython.

setup.py
```diff
@@ -0,0 +1,30 @@
+
+import os
+import sys
```
nigelsmall: Neither of these imports are used.

spacecowboy:
Fixed the setup script

PR #5, opened 2015-10-13
Fix the failure in CI

zhenlineo:
-o Fixed wrong input type for hex method
 -o Fixed wrong assertion on array

### DIFF
--- a/neo4j/connection.py
+++ b/neo4j/connection.py
@@ -20,7 +20,7 @@
 
 from __future__ import division
 
-from collections import deque, namedtuple
+from collections import deque
 from io import BytesIO
 import logging
 from os import environ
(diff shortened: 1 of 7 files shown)


neo4j/connection.py
```diff
@@ -70,7 +70,8 @@
 log_error = log.error
 
 
-def hex2(x):
+def hex2(str):
+    x = ord(str)
```
nigelsmall: Couple of things here: `str` is a reserved word in Python and this functionality varies between Python 2 and 3. Can pair on this with you :-)

test/session_test.py
```diff
@@ -190,6 +208,8 @@ def test_can_rollback_transaction(self):
 
             tx.rollback()
 
+            # I do not know how this framework works to close the tx after each method call
+            # But it should be closed due to the rollback
```
zhenlineo: My commit could go aways now

PR #14, opened 2015-11-24
Add versioning to API

pontusmelke:
Following the Java driver: changed `neo4j` -> `neo4j.v1` to
facilitate eventual API migration.

Built on top of https://github.com/neo4j/neo4j-python-driver/pull/13


test/test_packstream.py
```diff
@@ -19,13 +19,12 @@
...
 
+import struct
 from collections import OrderedDict
 from io import BytesIO
 from math import pi
-import struct
 from unittest import TestCase
-
-from neo4j.packstream import Packer, Unpacker, packb
+from neo4j.v1 import Packer, Unpacker, packb
```
nigelsmall: I think this should be `neo4j.v1.packstream`

neo4j/v1/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from .typesystem import *
+
+from .session import *
+
+from packstream import *
```
nigelsmall: No need to export this module at top level.

pontusmelke: It was done this way before, if I don't export here `test_packstream.py` is sad because it cannot import `Packer`, `Unpacker` and `packb` ?

PR #25, opened 2015-12-23
Added TCK for python

RagnarW:
TCK tests are set to run with runtests.sh

Python-drivers currently do not accept sending Path/Node/Relationship while Java-drivers do. So those tests are currently failing.

### DIFF
--- a/test/tck/configure_feature_files.py
+++ b/test/tck/configure_feature_files.py
@@ -1,0 +1,30 @@
+import os
+import tarfile
+import urllib2
+
+
+def clean_up():
+    dir_path = (os.path.dirname(os.path.realpath(__file__)))#'test/tck/'
+    files = os.listdir(dir_path)
+    for f in files:
+        if not os.path.isdir(f) and ".feature" in f:
+            os.remove(os.path.join(dir_path, f))
+
+
+def set_up():
+    dir_path = (os.path.dirname(os.path.realpath(__file__)))
+    feature_url = "https://s3-eu-west-1.amazonaws.com/remoting.neotechnology.com/driver-compliance/tck.tar.gz"
+    file_name = feature_url.split('/')[-1]
+    tar = open(file_name, 'w')
+    response = urllib2.urlopen(feature_url)
+    block_sz = 1024
+    while True:
+        buffer = response.read(block_sz)
+        if not buffer:
+            break
+        tar.write(buffer)
+    tar.close()
+    tar = tarfile.open(file_name)
+    tar.extractall(dir_path)
+    tar.close()
+    os.remove(file_name)
--- a/test/tck/configure_feature_files.py
+++ b/test/tck/configure_feature_files.py
@@ -1,7 +1,5 @@
 import os
 import tarfile
-import urllib2
-
 
 def clean_up():
     dir_path = (os.path.dirname(os.path.realpath(__file__)))
@@ -13,18 +11,28 @@ def clean_up():
 
 def set_up():
     dir_path = (os.path.dirname(os.path.realpath(__file__)))
-    feature_url = "https://s3-eu-west-1.amazonaws.com/remoting.neotechnology.com/driver-compliance/tck.tar.gz"
-    file_name = feature_url.split('/')[-1]
-    tar = open(file_name, 'w')
-    response = urllib2.urlopen(feature_url)
-    block_sz = 1024
-    while True:
-        buffer = response.read(block_sz)
-        if not buffer:
-            break
-        tar.write(buffer)
-    tar.close()
+    url = "https://s3-eu-west-1.amazonaws.com/remoting.neotechnology.com/driver-compliance/tck.tar.gz"
+    file_name = url.split('/')[-1]
+    _download_tar(url,file_name)
+
     tar = tarfile.open(file_name)
     tar.extractall(dir_path)
     tar.close()
     os.remove(file_name)
+
+
+def _download_tar(url, file_name):
+    try:
+        import urllib2
+        tar = open(file_name, 'w')
+        response = urllib2.urlopen(url)
+        block_sz = 1024
+        while True:
+            buffer = response.read(block_sz)
+            if not buffer:
+                break
+            tar.write(buffer)
+        tar.close()
+    except ImportError:
+        from urllib import request
+        request.urlretrieve(url, file_name)
--- a/test/tck/steps/bolt_type_steps.py
+++ b/test/tck/steps/bolt_type_steps.py
@@ -1,0 +1,241 @@
+from behave import *
+
+from test.tck import tck_util
+
+from neo4j.v1.typesystem import Node, Relationship, Path
+
+use_step_matcher("re")
+
+
+@given("A running database")
+def step_impl(context):
+    return None
+    # check if running
+
+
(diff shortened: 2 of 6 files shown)


test/tck/configure_feature_files.py
```diff
@@ -0,0 +1,30 @@
+import os
+import tarfile
+import urllib2
```
spacecowboy: Sorry about the random comment here, I just noted that this line will fail on Python3. urllib2 has been renamed urllib (and slightly restructured) there.

technige:
If I understand your description of this PR correctly, I would argue that the Java driver should _not_ permit Node/Rel/Path objects as parameters. Doing so is not supported by the current remoting interface (HTTP) and the behaviour for this is AFAIK not well-defined by Cypher, despite being permitted by the embedded API.

@jakewins: we had conversations about this earlier last year - has this changed?

test/tck/configure_feature_files.py
RagnarW: You are probably right. This has currently been build only focusing on python 2.7 at the moment.

test/tck/steps/bolt_type_steps.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@given("A running database")
+def step_impl(context):
+    return None
+    # check if running
+
+
+@given("a value (?P<Input>.+) of type (?P<BoltType>.+)")
+def step_impl(context, Input, BoltType):
```
technige: Any reason these arguments need to be TitleCased?

test/tck/steps/bolt_type_steps.py
```diff
@@ -0,0 +1,140 @@
...
+
+@step("adding map M to list L")
+def step_impl(context):
+    context.L.append(context.M)
+
+
+@when("adding a table of lists to the map M")
+def step_impl(context):
+    for row in context.table:
+        context.M['a' + str(len(context.M))] = tck_util.get_list_from_feature_file(row[1], row[0])
```
technige: Alternatively:

```
'a%d' % len(context.M)
```

test/tck/steps/bolt_type_steps.py
```diff
@@ -0,0 +1,140 @@
...
+def step_impl(context):
+    context.expected = context.M
+    context.results = {}
+    context.results["as_string"] = tck_util.send_string("RETURN " + tck_util.as_cypger_text(context.expected))
+    context.results["as_parameters"] = tck_util.send_parameters("RETURN {input}", {'input': context.expected})
+
+
+@then("the result returned from the server should be a single record with a single value")
+def step_impl(context):
+    assert len(context.results) > 0
```
technige: More Pythonic to simply write:

```
assert context.results
```

test/tck/steps/bolt_type_steps.py
```diff
@@ -0,0 +1,140 @@
...
+    for result in context.results.values():
+        result_value = result[0].values()[0]
+        assert result_value == context.expected
+        assert result_value.labels == context.expected.labels
+        assert result_value.keys() == context.expected.keys()
+        assert result_value.values() == context.expected.values()
+        assert result_value.items() == context.expected.items()
+        assert len(result_value) == len(context.expected)
+        assert iter(result_value) == iter(context.expected)
```
technige: Is `iter(...) == iter(...)` actually what you want to do here? the LHS and RHS will both return generators and this will compare the generator objects themselves rather than the contents.

You probably want to change `iter` to `list` which will consume the iterables and compare as lists instead.

test/tck/tck_util.py
```diff
@@ -0,0 +1,120 @@
...
+        return None
+    if type == 'Boolean':
+        return bool(value)
+    raise ValueError('No such type : %s', type)
+
+
+def as_cypger_text(expected):
+    if expected is None:
+        return "Null"
+    if isinstance(expected, unicode):
```
technige: This won't work across Python 2 and 3. Note how the `string` variable is generated here:
https://github.com/neo4j/neo4j-python-driver/blob/1.0/neo4j/v1/compat.py#L31

test/tck/tck_util.py
```diff
@@ -0,0 +1,120 @@
...
+    if type == 'String':
+        return value
+    if type == 'Null':
+        return None
+    if type == 'Boolean':
+        return bool(value)
+    raise ValueError('No such type : %s', type)
+
+
+def as_cypger_text(expected):
```
technige: Typo: "as_cypher_text"

test/tck/tck_util.py
```diff
@@ -0,0 +1,120 @@
...
+        return int(value)
+    if type == 'Float':
+        return float(value)
+    if type == 'String':
+        return value
+    if type == 'Null':
+        return None
+    if type == 'Boolean':
+        return bool(value)
+    raise ValueError('No such type : %s', type)
```
technige:
```
raise ValueError('No such type : %s' % type)
```

test/tck/tck_util.py
```diff
@@ -0,0 +1,120 @@
...
+    return result
+
+
+def get_bolt_value(type, value):
+    if type == 'Integer':
+        return int(value)
+    if type == 'Float':
+        return float(value)
+    if type == 'String':
+        return value
```
technige: How can you be sure that `value` is always passed in as a string?

test/tck/tck_util.py
```diff
@@ -0,0 +1,120 @@
...
+        fu = get_random_string
+        args = [3]
+    elif type == 'Null':
+        fu = get_none
+        args = []
+    elif type == 'Boolean':
+        fu = get_random_bool
+        args = []
+    else:
+        raise ValueError('No such type : %s', type)
```
technige:
```
raise ValueError('No such type : %s' % type)
```

test/tck/tck_util.py
```diff
@@ -0,0 +1,120 @@
...
+    fu, args = _get_random_func(type)
+    return [fu(*args) for _ in range(size)]
+
+
+def get_dict_of_random_type(size, type):
+    fu, args = _get_random_func(type)
+    map = {}
+    for i in range(size):
+        map['a' + str(i)] = fu(*args)
+    return map
```
technige: Simpler to use a dictionary comprehension here:

```
return {'a%d' % i: fu(*args) for i in range(size)}
```

test/tck/steps/bolt_type_steps.py
RagnarW: Nope. They just have that format in the feature file

test/tck/steps/bolt_type_steps.py
RagnarW: This function should actually have been removed in the last commit since we do not send nodes as parameters any more. But thanks for the feedback anyway.

test/tck/tck_util.py
spacecowboy: Dictionary comprehensions are not available in Python2 though

RagnarW: Yes, it works fine on 2.7.

spacecowboy: Oh, you're right. I was mistaken here. Carry on! :)

PR #29, opened 2016-01-11
1.0 session pool

technige:


neo4j/v1/session.py
```diff
@@ -341,6 +355,11 @@ def __enter__(self):
     def __exit__(self, exc_type, exc_value, traceback):
         self.close()
 
+    def reset(self):
```
jakewins: Should reset really be a user-facing feature? I'd argue we should encourage users of the Session API to simply close sessions if they are unclear about their state, unless there's a good use case that warrants doing this instead?

jakewins:
This is brilliant, super happy to see this come together. I'd love a test that ensures the `defunct` functionality works - perhaps one that checks that we can handle a fatal error occurring on the connection  by stubbing out the underlying channel or something, and one that checks that the pool recovers if the server dies?

PR #31, opened 2016-01-19
Cursor API

technige:
This provides manual and yield-based cursor navigation capabilities but does not (yet) include LINQ-like methods (`first`, `single`, etc) or `peek`. As this is a big change already, I would prefer to add those as part of a separate PR if indeed we agree they are required for the next milestone.

This PR also updates the examples for the docs and README to reflect the new cursor-based approach and removes the old benchmarking code that is no longer consistent, given so much is now done lazily. We should rethink benchmarking.


README.rst
```diff
@@ -28,8 +28,10 @@ Example Usage
     driver = GraphDatabase.driver("bolt://localhost")
     session = driver.session()
     session.run("CREATE (a:Person {name:'Bob'})")
-    for name, in session.run("MATCH (a:Person) RETURN a.name AS name"):
-        print(name)
+    cursor = session.run("MATCH (a:Person) RETURN a.name AS name")
+    while cursor.next()
```
jakewins, line 32: missing colon at end of this line

neo4j/v1/session.py
```diff
@@ -120,7 +102,7 @@ def session(self):
         return session
 
     def recycle(self, session):
```
jakewins, line 104: Unrelated to this PR - but how come this is a public method, rather than having `Session` return itself to the pool on `close`? Doing it on close is usually helpful, since then you only have to have the session to recycle it, rather than also pass around the graph database instance. 

Also a bit surprised it's on GraphDatabase, rather than on Driver - does that mean Driver instances share access to a process-global pool?

neo4j/v1/session.py
```diff
@@ -143,60 +125,124 @@ class Result(list):
...
+
+    def is_open(self):
+        """ Return ``True`` if this cursor is still open, ``False`` otherwise.
+        """
+        return bool(self._connection)
 
-    def on_header(self, metadata):
-        """ Called on receipt of the result header.
+    def close(self):
+        """ Consume the remainder of this result and detach the connection
```
jakewins, line 146: To a novice user, not sure what "detach the connection" means - what does that mean for how I write my code?

neo4j/v1/session.py
jakewins, line 104: Realized that session.close already does this :) Still unsure if this should be public and why it's on GraphDatabase, but again, unrelated to this PR.

PR #39, opened 2016-03-03
1.0 examples

technige:
Aligned TLS settings with Java driver and completed examples.


examples/test_examples.py
```diff
@@ -19,18 +19,26 @@
...
 
 
+from unittest import skip
+
+from neo4j.v1 import basic_auth, TRUST_ON_FIRST_USE, TRUST_SIGNED_CERTIFICATES
 from test.util import ServerTestCase
 
+# Do not change the contents of this tagged section without good reason*
 # tag::minimal-example-import[]
-from neo4j.v1 import GraphDatabase, basic_auth
```
jakewins, line 25: Shouldn't this include `basic_auth` though? For 99% of first-time users, this import line is what they should use.

examples/test_examples.py
```diff
@@ -68,24 +76,24 @@ def test_configuration(self):
...
-        # TODO: Unfortunately, this feature is not yet implemented for Python
-        pass
+        driver = GraphDatabase.driver("bolt://localhost", auth=auth_token, encrypted=True)
         # end::tls-require-encryption[]
 
     def test_tls_trust_on_first_use(self):
         # tag::tls-trust-on-first-use[]
-        # TODO: Unfortunately, this feature is not yet implemented for Python
-        pass
+        driver = GraphDatabase.driver("bolt://localhost", auth=auth_token, encrypted=True, trust=TRUST_ON_FIRST_USE)
```
jakewins: Same here - IMO we should prefer to show users a working example, they will not know what `auth_token` is here.

examples/test_examples.py
```diff
@@ -68,24 +76,24 @@ def test_configuration(self):
...
         # end::tls-trust-on-first-use[]
+        assert driver
 
+    @skip("testing verified certificates not yet supported ")
```
jakewins, line 88: This makes me uncomfortable - if we don't have tests for a feature, I don't think we should advertise it in the documentation. Better say we don't yet have it until it is testable?

jakewins:
Nice! See comments inline, nothing major - but I think we should keep showing the full `basic_auth(..)` in the examples, and I think we should wait to document trusted certificates until we have a test ensuring the feature works and keeps working.

examples/test_examples.py
technige, line 25: You're right, I'll fix that.

examples/test_examples.py
technige: IIRC the manual displays the imports specifically in their own section since they apply to all examples. So fixing that in the case above fixes it for all examples.

examples/test_examples.py
technige, line 88: So you're suggesting to remove the test for the time being? I saw this as a handy reminder for us to complete the test when we have a good way to test it. Bearing in mind that test code doesn't get bundled with the published driver.

PR #41, opened 2016-03-04
1.0 tck tests - ResultSummary API and Equality

RagnarW:
The follow needs to be fixed before merge:

Equality:
- ~~Decide what equality means. The feature file is designed for using IDs~~
- ~~Should Path direction matter? The feature file is designed to test direction~~
- ~~Should JS graph-objects have equals. There is no equality in JS should this be implemented?~~

ResultApi:
- ~~ProfiledPlan: should the returned rows/records return "records" or "rows"~~
- ~~Statistics: should statics be specified and should update statistics be part of it?~~


test/test_session.py
```diff
@@ -335,15 +335,15 @@ def test_automatic_reset_after_failure(self):
...
-    def test_defunct(self):
-        from neo4j.v1.connection import ChunkChannel, ProtocolError
-        with GraphDatabase.driver("bolt://localhost", auth=auth_token).session() as session:
-            assert not session.connection.defunct
-            with patch.object(ChunkChannel, "chunk_reader", side_effect=ProtocolError()):
-                with self.assertRaises(ProtocolError):
-                    session.run("RETURN 1").consume()
-            assert session.connection.defunct
-            assert session.connection.closed
+    # def test_defunct(self):
```
pontusmelke: ?

PR #43, opened 2016-03-13
allow basic authentication username/password to be provided as part o…

karabijavad:
…f url

### DIFF
--- a/neo4j/v1/session.py
+++ b/neo4j/v1/session.py
@@ -94,6 +94,8 @@ class Driver(object):
         if parsed.scheme == "bolt":
             self.host = parsed.hostname
             self.port = parsed.port
+            if parsed.username:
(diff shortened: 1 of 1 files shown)


technige:
We have previously discussed this and decided against allowing it for any of our drivers. For our reasoning, please refer to RFC 3986 section 3.2.1. Specifically:

```
Use of the format "user:password" in the userinfo field is deprecated. Applications should not render as clear text any data after the first colon (":") character found within a userinfo subcomponent unless the data after the colon is the empty string (indicating no password). Applications may choose to ignore or reject such data when it is received as part of a reference and should reject the storage of such data in unencrypted form. The passing of authentication information in clear text has proven to be a security risk in almost every case where it has been used.
```

karabijavad:
ah, sorry @nigelsmall 
thanks!

PR #45, opened 2016-03-18
Rename identity to id

zhenlineo:


neo4j/v1/types.py
```diff
@@ -140,21 +139,21 @@ class UnboundRelationship(BaseRelationship):
     """
 
     @classmethod
-    def hydrate(cls, identity, type, properties=None):
+    def hydrate(cls, id, type, properties=None):
```
technige: The name `id` refers to a built-in function in Python so using that name as a variable shadows that built-in. Standard practice would be to use `id_` instead. This doesn't affect member attributes of objects, i.e. `alice.id`.

Issue #46, opened 2016-04-01
SSL import issue in Python 2.7 earlier than 2.7.9


technige:
For details, see https://github.com/nigelsmall/py2neo/issues/487

technige:
Likely fix is to handle `ImportError` gracefully and disable SSL support in 2.7 if below 2.7.9.

Alan-R:
I can live with this. Is this where I should have opened the issue?

technige:
No, that's fine. You use py2neo and used the library indirectly from there so you raised it in the right place.

technige:
This was addressed in 92e47d416481a9ef2165d547c01cdf3496ac99e1

PR #48, opened 2016-04-04
Graceful SSL degradation

technige:
This should fix issue #46

### DIFF
--- a/examples/test_examples.py
+++ b/examples/test_examples.py
@@ -19,9 +19,9 @@
 # limitations under the License.
 
 
-from unittest import skip
+from unittest import skip, skipUnless
 
-from neo4j.v1 import TRUST_ON_FIRST_USE, TRUST_SIGNED_CERTIFICATES
+from neo4j.v1 import TRUST_ON_FIRST_USE, TRUST_SIGNED_CERTIFICATES, SSL_AVAILABLE
 from test.util import ServerTestCase
 
 # Do not change the contents of this tagged section without good reason*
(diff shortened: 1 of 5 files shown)


pontusmelke:
Looks fine to me. Just a question: Could it be a problem that we are silently _downgrading_ default to be insecure. I am thinking could we give some sort of warning when running the driver on older version, maybe here https://github.com/neo4j/neo4j-python-driver/blob/1.0-graceful-ssl-degradation/neo4j/v1/ssl_compat.py#L24?

technige:
Yes, I wondered that too but wasn't sure where/when to inject the warning. Thinking about this again though, maybe a warning should appear on the _firs_t time that a driver instance is created with default encryption/trust.

Alan-R:
Is it possible for the application to know what has happened one way or the other?

Can you disable non-SSL (http-only) in neo4j if you really care... That would apply to all uses of the API, which would be broader and more effective...

zhenlineo:
I am not sure if I got your question, but I'll give it a try. 
You could config the database to accept SSL enabled, http-only connections.

Say you has a port 1234, you could config it to open for ssl-enabled, http-only connections by using the following settings in a Neo4j database:

```
dbms.connector.1234.address=localhost:1234
dbms.connector.1234.enabled=true
dbms.connector.1234.encryption=TLS
dbms.connector.1234.type=HTTP
```

If you got another port 4321, and you want to config it differently, then just set it accordingly, such as:

```
dbms.connector.4321.address=localhost:4321
dbms.connector.4321.enabled=true
dbms.connector.4321.encryption=REQUIRED
dbms.connector.4321.type=BOLT
```

Then you explicitly set what kind of connections that you want for different ports.
For more info: http://neo4j.com/docs/operations-manual/3.0/#_configuring_bolt_connectors

Hope this answered your questions :)

Alan-R:
@nigelsmall - I might recommend that people do things like @zhenlineo suggested if they really care about making sure that they don't accidentally get unencrypted connections. That's better than anything you can do through neo4j anyway. But having a way to ask "Did I get an encrypted connection?" still seems like a reasonable thought. Not required, just food for thought. If your application cares, then you can ask. If you don't know about SSL (too old), then your current action is backwards-compatible.

@zhenlineo It looks like you can't disable http-only connections (_exactly one_ http connector).

technige:
You can already work this out by using `driver.encryption` which will return `True` or `False` depending on whether the driver produces encrypted sessions or not.

Alan-R:
Good answer. Thanks @nigelsmall !

PR #51, opened 2016-04-07
1.0 tck tests - Error handling

RagnarW:
Adds TCK:
- Error reporting
- StatementResult navigation
- Auth

Fixes in driver code:
- Cast exception if Session already has active transaction when using 'run' or 'beginTransaction'
- Clears transaction from session when recycled 
- Changed StatementResult to return tuple of keys, same format as record.
- Some changes in error messages and types.


neo4j/v1/connection.py
```diff
@@ -379,7 +379,7 @@ def connect(host, port=None, ssl_context=None, **config):
     try:
         s = create_connection((host, port))
     except SocketError as error:
-        if error.errno == 111:
+        if error.errno == 111 or error.errno == 61:
```
technige, line 386: A comment might be useful here to highlight what errors these are.

neo4j/v1/session.py
```diff
@@ -103,11 +103,13 @@ class Driver(object):
...
         parsed = urlparse(self.url)
-        if parsed.scheme == "bolt":
+        transports = ['bolt']
+        if parsed.scheme in transports:
             self.host = parsed.hostname
             self.port = parsed.port
         else:
-            raise ValueError("Unsupported URL scheme: %s" % parsed.scheme)
+            raise ProtocolError("Unsupported transport: '%s' in url: '%s'. Supported transports are: '%s'." %
+                                (parsed.scheme, url, transports))
```
technige: Instead of generalising this, can we just hard-code "Bolt" into the error message. We can generalise when we support more than one protocol. Also, we should use the term "URI scheme" instead of "transport".

neo4j/v1/session.py
```diff
@@ -103,11 +103,13 @@ class Driver(object):
     def __init__(self, url, **config):
         self.url = url
         parsed = urlparse(self.url)
-        if parsed.scheme == "bolt":
+        transports = ['bolt']
```
technige: A list here should not be necessary (see note below)

neo4j/v1/session.py
```diff
@@ -239,7 +241,7 @@ def keys(self):
         # Fetch messages until we have the header or a failure
         while self._keys is None and not self._consumed:
             self.connection.fetch()
-        return self._keys
+        return tuple(self._keys)
```
technige, line 243: Was this previously coming through as a non-tuple?

neo4j/v1/session.py
```diff
@@ -477,6 +478,12 @@ def healthy(self):
         return None if connection.closed else not connection.defunct
 
     def run(self, statement, parameters=None):
+        if self.transaction:
+            raise ProtocolError("Please close the currently open transaction object before running more "
+                                "statements/transactions in the current session.")
```
technige: Instead of suggesting the transaction be closed as the only course of action, we should say something like:

> Statements cannot be run directly on a session with an open transaction; either run from within the transaction or use a different session.

neo4j/v1/session.py
```diff
@@ -477,6 +478,12 @@ def healthy(self):
...
+        return self._run(statement, parameters)
+
+    def _run(self, statement, parameters=None):
         """ Run a parameterised Cypher statement.
 
         :param statement: Cypher statement to execute
```
technige, line 482: This comment should be moved to `run`.

neo4j/v1/session.py
```diff
@@ -517,14 +524,18 @@ def close(self):
...
 
     def begin_transaction(self):
         """ Create a new :class:`.Transaction` within this session.
 
         :return: new :class:`.Transaction` instance.
         """
-        assert not self.transaction
+        if self.transaction:
+            raise ProtocolError("Please close the currently open transaction object before running more "
+                                "statements/transactions in the current session.")
```
technige: Ditto message in `run`.

neo4j/v1/session.py
RagnarW: Sure. These changes were made to fit how the Java Driver did. I have no preference here.

technige: :+1:

neo4j/v1/session.py
RagnarW: I am fine with that. Same as before. This is taken from the Java Driver since I figured it made sense to have similar error messages for the same errors

neo4j/v1/session.py
RagnarW, line 243: it comes as a list

technige, line 243: Then :+1:

PR #52, opened 2016-04-08
Peek and single methods

technige:


neo4j/v1/session.py
```diff
@@ -256,6 +256,32 @@ def consume(self):
...
 
+    def single(self):
+        """ Return the next record, failing if none or more than one remain.
+        """
+        records = list(self)
+        num_records = len(records)
+        if num_records == 0:
+            raise ResultError("No records found in stream")
+        elif num_records != 1:
+            raise ResultError("Multiple records found in stream")
```
zhenlineo, line 267: _single_ - **consume all remaining records** and return the first of these, failing if there is not exactly one record
According to the [card](https://trello.com/c/UYYMVLGj/884-clarify-how-calling-methods-on-statement-result-changes-the-state-of-the-result)

PR #55, opened 2016-04-19
Added three missing examples

zhenlineo:
Recreate this pr to let PR build pick it up


zhenlineo:
I am merging this in as I cannot see the reason why it is red. I am guessing it is because the PR build is set wrongly. So let's do this experiment by merging it in.

PR #59, opened 2016-05-05
update README to demonstrate basic_auth usage

stevenpollack:
- example didn't work without explicitly setting up authentication.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,8 +24,8 @@ Example Usage
 
 .. code:: python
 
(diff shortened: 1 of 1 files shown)


zhenlineo:
Thanks so much for the change.
Would you mind to sign CLA (http://neo4j.com/developer/cla/) so that we could accept you PR?

zhenlineo:
Thanks for singing the CLA

PR #91, opened 2016-09-18
Core ORM components for Cypher code generation

dgabriele:
Signed-off-by: Daniel Gabriele d.gabri3le@gmail.com


technige:
Hi @dgabriele 

Thank you for your submission and for putting so much effort into your code! However, ORM/OGM isn't part of our roadmap for the official drivers yet and it's something that we'd want to look at in a uniform way across languages. Could I suggest that you create this as a new project that depends on the driver? We'll be happy to host this in a `neo4j-contrib` repository if you'd like.

Kind regards
Nigel

PR #104, opened 2016-12-05
Bookmarking

technige:


neo4j/v1/session.py
```diff
@@ -336,6 +338,8 @@ def run(self, statement, parameters=None, **kwparameters):
         :return: Cypher result
         :rtype: :class:`.StatementResult`
         """
+        self.last_bookmark = None
```
lutovich, line 270: Do we need to do this initialization together with initialization in line 314?

neo4j/v1/session.py
```diff
@@ -387,15 +391,23 @@ def begin_transaction(self):
...
+            parameters["bookmark"] = bookmark
+
+        self.run("BEGIN", parameters)
         self.transaction = Transaction(self, on_close=clear_transaction)
         return self.transaction
 
     def commit_transaction(self):
-        self.run("COMMIT")
+        result = self.run("COMMIT")
+        self.connection.sync()
```
lutovich, line 335: Is there a way to test that we send commit and rollback messages eagerly?

neo4j/v1/bolt.py
```diff
@@ -238,19 +243,21 @@ def __init__(self, sock, **config):
         # Pick up the server certificate, if any
         self.der_encoded_server_certificate = config.get("der_encoded_server_certificate")
 
+        def on_success(metadata):
+            self.server_version = metadata.get("server")
```
lutovich, line 255: `server_version` field is initialized here but where is it read?

neo4j/v1/session.py
```diff
@@ -372,11 +376,11 @@ def close(self):
...
         if self.connection:
             if not self.connection.closed:
-                self.connection.fetch_all()
+                self.connection.sync()
             self.connection.in_use = False
             self.connection = None
 
-    def begin_transaction(self):
+    def begin_transaction(self, bookmark=None):
         """ Create a new :class:`.Transaction` within this session.
```
lutovich, line 313: Should docstring also say something about the bookmark parameter?

lutovich (commented):
@nigelsmall review completed, couple small comments.

neo4j/v1/session.py
technige: Ah, yes we do. This clears the bookmark before a run, it's not initialisation.

neo4j/v1/session.py
technige: Tests added for that.

neo4j/v1/bolt.py
technige: Just used in a test right now. It will conflict with #108 so I'll adjust it properly for that.

Issue #84, opened 2016-06-28
Documentation example doesn't work


PabloCastellano:
From https://neo4j.com/docs/api/python-driver/current/#example

This line:

```
>>> for friend, in session.run("MATCH (a:Person {name:'Alice'})-[:KNOWS]->(x) RETURN x"):
...     print('Alice says, "hello, %s"' % friend["name"])
... 
Traceback (most recent call last):
  File "<input>", line 2, in <module>
    print('Alice says, "hello, %s"' % friend["name"])
TypeError: string indices must be integers
```

EmilStenstrom:
This problem is likely just an extra comma after `friend`: 

``` python
for friend, in session.run(...)
```

liutec:
The comma is not extra. 

``` python
for friend, in session.run("MATCH (a:Person {name:'Alice'})-[:KNOWS]->(x) RETURN x"):
    print('Alice says, "hello, %s"' % friend["name"])
```

is equivalent to:

``` python
for record in session.run("MATCH (a:Person {name:'Alice'})-[:KNOWS]->(x) RETURN a, x"):
    friend, = record
    print('Alice says, "hello, %s"' % friend["name"])
```

`Record.__iter__` will be called in order to assign the first returned value (`x`) to `friend`.

It's use is to extract returned values like so:

``` python
for a, x in session.run("MATCH (a:Person {name:'Alice'})-[:KNOWS]->(x) RETURN a,x"):
    print('%s says, "hello, %s"' % (a["name"], x["name"]))
```

instead of having to address them explicitly in the record:

``` python
for record in session.run("MATCH (a:Person {name:'Alice'})-[:KNOWS]->(x) RETURN a,x"):
    print('%s says, "hello, %s"' % (record["a"]["name"], record["x"]["name"]))
```

The problem is that `Record.__iter__` returns `iter(self._keys)` instead of `iter(self._values)`.

I believe this is a flaw in the implementation of the `Record` class.
https://github.com/neo4j/neo4j-python-driver/blob/f7c2f771c4b55ef6b81b38035e091c406a8aba46/neo4j/v1/session.py#L477

I may be wrong, I just started to evaluate this library, but I see no point in providing iteration support to fetch the already known variable names instead of the actual values.

technige:
Fixed in #111

PR #98, opened 2016-11-23
Reworked the docs example code

petraselmer:

PR #111, opened 2016-12-14
Fix example code

technige:


technige:
Supersedes #98

technige:
Fixes #84

PR #122, opened 2017-01-11
Move Unauthorized into public API

technige:


zhenlineo:
Will this Unauthorized error also be used for client failed to auth server, a.k.a. the server's cert is not trusted?

technige:
It's not currently. The built-in SSL exceptions are used there. I think the two cases are probably better distinct as they require separate actions to be taken.

zhenlineo:
Could you drop a trello card for the error type changes in python driver so that other drivers could also take action and catch-up later? Then you are free to merge these changes.

PR #130, opened 2017-01-23
Get the correct socket infor in case of ipv6.

dmanchon:


technige:
That was quick, thank you :-)

My thoughts on the code:
1. We should probably wrap this up into a small helper function to reduce duplication: `get_host_port(socket) -> host, port`.
2. Explicit checks on `AF_INET`, rather than a simple `else` would ensure this can't happen again. If the address family isn't recognised at all, we should raise a `ProtocolError`.
3. Tests.

Have you signed our [CLA](https://neo4j.com/developer/cla/)? I'm happy to work on this otherwise.

dmanchon:
Agree with all. I still didnt figure out yet how correctly run the tests.

Thanks.

technige:
`./runtests.sh` should do it though you'll need to `pip install boltkit -r test/requirements.txt` first.

dmanchon:
Thanks. I just agreed the CLA and i the tests passed in my local machine but unfortunately they are failing in the CI

technige:
Closing this PR as this has now been added in https://github.com/neo4j/neo4j-python-driver/commit/6c5f6fb2e67c57ebdcb520e3ced38e180c476229

PR #148, opened 2017-03-30
Try all resolved addresses for direct connection

technige:


jexp (approved):
Works now with the default Neo4j server install and localhost. Thanks a lot for the quick fix.

PR #155, opened 2017-04-24
Roll back to initial server if running out of routers

zhenlineo:


neo4j/v1/routing.py
```diff
@@ -216,16 +217,32 @@ def fetch_routing_table(self, address):
...
+            return
+
+        initial_routers = resolve(self.initial_address)
         for router in copy_of_routers:
-            new_routing_table = self.fetch_routing_table(router)
-            if new_routing_table is not None:
-                self.routing_table.update(new_routing_table)
+            if router in initial_routers:
+                initial_routers.remove(router)
+        if len(initial_routers) != 0:
```
zhenlineo: if initial_routers:

PR #156, opened 2017-04-24
Support transient but not terminated by user errors in retry

zhenlineo:


neo4j/v1/api.py
```diff
@@ -461,6 +463,17 @@ def __bookmark__(self, result):
         pass
 
 
+def is_retriable_transientError(error):
+    """
+    :type error: TransientError
+    """
+    if (error.code != "Neo.TransientError.Transaction.Terminated"
```
zhenlineo: return error.code in ("error1", "error2")

PR #160, opened 2017-04-27
Fix examples

zhenlineo:


zhenlineo:
@srbaker It seems the import will fail on 3.4 due to missing path. I think this should fix the problem

PR #169, opened 2017-05-30
Bytes

technige:
Support for byte arrays.


tardyp:
thanks! the patch looks good. even if using <3.2, we know that we messed up the decoding

PR #176, opened 2017-07-20
Routing driver forgets addresses on some errors

lutovich:
This PR makes routing driver forget addresses and purge their connections on network and database errors. It also makes driver forget write server addresses when they fail as writers, connections remain in the pool because such machines can remain readers/routers.

### DIFF
--- a/neo4j/bolt/connection.py
+++ b/neo4j/bolt/connection.py
@@ -144,6 +144,9 @@ class Connection(object):
     #: Error class used for raising connection errors
     Error = ServiceUnavailable
 
+    #: The function to handle send and receive errors
+    error_handler = None
+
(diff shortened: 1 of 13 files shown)


zhenlineo:
LGTM. CC @technige

technige (changes requested):
Not keen on the `error_handler` changes to `connection.py`. It's quite unusual to dynamically attach methods in this way and the new `try...except` blocks catch all `Exception` types, which is generally considered too brute force.

I'm happy to pair on this.

lutovich:
@technige dynamic `error_handler` is now replaced by constructor injection of `ConnectionErrorHandler`. Later is different for direct and routing driver.

technige (approved):

PR #197, opened 2017-11-23
Do not purge inUse connections

zhenlineo:
Do not purge inUse connections when connection error happens or new routing table is available

Instead we deactive the server (a.k.a. closing all idle connections) in connection pool.

When an connection failure happens, the connection server will be removed from the routing table and all idle connections with this server will be closed in the connection pool.

When a new routing table is available, all idle connections towards the servers that are not in the new routing table will be removed. If the server has no inUse connections, then this server will also be totally removed from the connection pool.

Note: 
* There is no extra protection to against `acquire` to create a new connection with an inactive server except that the server should not be in the routing table and therefore should not be used to `acquire` new connections.
* When error happens on a connection, we will deactivate the server but we will probably not remove all server's connections from connection pool as the failed connection is highly still inUse by the broken session.

### DIFF
--- a/neo4j/bolt/connection.py
+++ b/neo4j/bolt/connection.py
@@ -492,6 +492,25 @@ class ConnectionPool(object):
         else:
             return sum(1 if connection.in_use else 0 for connection in connections)
 
+    def deactivate(self, address):
+        """ Deactivate an address from the connection pool, if present, closing
+        all idle connection to that address
+        """
+        with self.lock:
+            try:
+                connections = self.connections[address]
+            except KeyError: # already removed from the connection pool
+                return
+            for conn in list(connections):
+                if not conn.in_use:
+                    connections.remove(conn)
+                    try:
+                        conn.close()
+                    except IOError:
+                        pass
+            if len(connections) == 0:
+                self.remove(address)
+
     def remove(self, address):
         """ Remove an address from the connection pool, if present, closing
         all connections to that address.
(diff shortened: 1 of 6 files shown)


neo4j/v1/routing.py
```diff
@@ -140,6 +143,9 @@ def update(self, new_routing_table):
         self.last_updated_time = self.timer()
         self.ttl = new_routing_table.ttl
 
+    def servers(self):
+        return set(self.routers.elements() + self.writers.elements() + self.readers.elements())
```
technige: This can just be `return set(self.routers) | set(self.writers) | set(self.readers)` and there is no need to add an `elements` accessor on `OrderedSet`.

neo4j/v1/routing.py
```diff
@@ -81,6 +81,9 @@ def replace(self, elements=()):
         e.clear()
         e.update(OrderedDict.fromkeys(elements))
 
+    def elements(self):
+        return list(self._elements)
```
technige: Not needed - see below.

neo4j/bolt/connection.py
```diff
@@ -492,6 +492,25 @@ def in_use_connection_count(self, address):
...
+            except KeyError: # already removed from the connection pool
+                return
+            for conn in list(connections):
+                if not conn.in_use:
+                    connections.remove(conn)
+                    try:
+                        conn.close()
+                    except IOError:
+                        pass
+            if len(connections) == 0:
```
technige: `if not connections`

Issue #201, opened 2017-11-29
Input buffer capacity exceeded when reading large values


liutec:
When reading results with values larger than the ChunkedInputBuffer's default capacity, a non-descriptive error gets triggered.

It would be nice to have the buffer capacities configurable along with a specific error message.

I'm currently using this as a workaround:

```python
class LargeChunkedInputBuffer(ChunkedInputBuffer):
    def __init__(self, capacity=4194304):
        super().__init__(capacity)

neo4j.bolt.connection.ChunkedInputBuffer = LargeChunkedInputBuffer
```

I realize it's not the best idea to store large values in Neo4j but as long as it's not prohibited they should be readable.

https://github.com/liutec/neo4j-python-driver/commit/165946f22af4a22c2487eeab74b953d752e6ac05

zhenlineo:
HI @liutec,

Thanks for bring this issue to us. It should not be needed to enlarge the capacity manually as the buffer should automatically grow for big records. Could you share us what error did you got?

Cheers,
Zhen

liutec:
Hello,

sorry for the late reply as well as for not including all the steps to reproduce this issue.
I was under the impression that the buffer has a fixed maximum size to avoid re-allocations. If the buffer was meant to grow beyond the capacity then it's clearly a problem only on Python 3 https://bugs.python.org/issue29178 

neo4j server  v3.0.5
Python 3.5.2 
neo4j-driver >= 1.4.0 (did not test on older versions)

```python
import sys
import neo4j
from neo4j.v1 import GraphDatabase


print(sys.version_info)
print("neo4j-driver", neo4j.__version__)

neo4j_driver = GraphDatabase.driver(
    uri="bolt://xxx.xxx.xxx.xxx:7687",
    auth=("neo4j", "neo4j")
)

cypher = "MERGE (n:TestNode) SET n.data = $data"
params = {
    "data": "x" * 524111  # does not reproduce below 524111
}

with neo4j_driver.session() as tx:
    tx.run(cypher, params)
del params

cypher = "MATCH (n:TestNode) RETURN n.data AS data LIMIT 1"
with neo4j_driver.session() as tx:
    records = tx.run(cypher)
    for record in records:
        print(len(record['data']))
```

```
sys.version_info(major=3, minor=5, micro=2, releaselevel='final', serial=0)
neo4j-driver 1.4.0
Exception ignored in: 'neo4j.bolt._io.ChunkedInputBuffer.receive'
[...]

Issue #165, opened 2017-05-04
python 3 return single collection of JSONs


tomasonjo:
I have this query i am running:

from neo4j.v1 import GraphDatabase

```
driver = GraphDatabase.driver("bolt://localhost:7687", auth=("neo4j", "neo4j"))


session = driver.session()


result = session.run("MATCH (a)-[r]->(b) WITH collect({source: id(a),target: id(b),caption: type(r)}) AS edges RETURN edges")

for record in result:
    print(record["edges"])
```

Now this works fine on my MAC, which uses python 2.7, but on my Windows Laptop, where I have Anaconda3 and python3 i get the following error:

> File "C:\ProgramData\Anaconda3\lib\site-packages\neo4j\v1\api.py", line 663, i
> n records
>     self._session.fetch()
>   File "C:\ProgramData\Anaconda3\lib\site-packages\neo4j\v1\api.py", line 321, i
> n fetch
>     detail_count, _ = self._connection.fetch()
>   File "C:\ProgramData\Anaconda3\lib\site-packages\neo4j\bolt\connection.py", li
> ne 255, in fetch
>     self._receive()
>   File "C:\ProgramData\Anaconda3\lib\site-packages\neo4j\bolt\connection.py", li
> ne 284, in _receive
>     received = self.input_buffer.receive_message(self.socket, 8192)
>   File "C:\ProgramData\Anaconda3\lib\site-packages\neo4j\bolt\io.py", line 155,
> in receive_message
>     received = self.receive(socket, n)
>   File "C:\ProgramData\Anaconda3\lib\site-packages\neo4j\bolt\io.py", line 139,
> in receive
>     self._data[self._extent:new_extent] = data
> BufferError: Existing exports of data: object cannot be re-sized

zhenlineo:
Could you provide the driver and server version that you are seeing this issue?

tomasonjo:
server version is 3.1.3, will upload the driver version when I get home

tomasonjo:
driver version is 1.2.1

john-bodley:
@zhenlineo I've seen a similar issue when trying to fetch entities from the graph, 

```python
Exception BufferError: 'Existing exports of data: object cannot be re-sized' in 'neo4j.bolt._io.ChunkedInputBuffer.receive' ignored
```

[...]

PR #204, opened 2017-12-04
Read large values stress test - nested reference buffer bug #201 #165

liutec:
This test should fail because a reference to the previous _view is kept in [ChunkedInputBuffer.frame_message](https://github.com/neo4j/neo4j-python-driver/blob/1.5/neo4j/bolt/io.py#L194)
Fix to follow.

### DIFF
--- a/neo4j/bolt/io.py
+++ b/neo4j/bolt/io.py
@@ -64,24 +64,31 @@ class MessageFrame(object):
     def read(self, n):
         if n == 0 or self._current_pane == -1:
             return _empty_view
-        p, q = self._panes[self._current_pane]
-        size = q - p
-        remaining = size - self._current_offset
-        if n <= remaining:
+        value = bytearray(b"\x00" * n)
+        value_end = 0
+        while n > 0 and self._current_pane >= 0:
+            p, q = self._panes[self._current_pane]
+            size = q - p
+            remaining = size - self._current_offset
             start = p + self._current_offset
-            end = start + n
-            if n < remaining:
-                self._current_offset += n
+            if n <= remaining:
+                end = start + n
+                if n < remaining:
+                    self._current_offset += n
+                else:
+                    self._next_pane()
             else:
+                end = q
                 self._next_pane()
-            return memoryview(self._view[start:end])
-        start = p + self._current_offset
-        end = q
-        value = bytearray(self._view[start:end])
-        self._next_pane()
-        if len(value) < n and self._current_pane >= 0:
-            value.extend(self.read(n - (end - start)))
-        return value
+            part_size = end - start
+            next_value_end = value_end + part_size
+            value[value_end:next_value_end] = self._view[start:end]
+            n -= part_size
+            value_end = next_value_end
+            continue
+        if n > 0:
+            value = value[:value_end]
+        return memoryview(value)
 
 
 class ChunkedInputBuffer(object):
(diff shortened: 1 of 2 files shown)


zhenlineo:
Hi @liutec,

Sorry for the late reply. I included your improvement to remove the recursion in #206 together with the fix for releasing the reference of view. Your fix to change to `_data` also works, but our initial plan was to use view to just expose part of the `_data` array, therefore we still prefer to use the `close` method to release the reference instead.

Also this part of the code also need to compile to C, therefore I started a PR to address some changes to Cython code. 

Would you also try out the changes in #206 and see if that also works for you? I failed to make client throw OOM before server. Also I change a bit in your `ChunkOutput#write` code. I am super bad at managing buffers. It would be super if you could review and test out.

Finally if it is okay, would you mind sign our [CLA](https://neo4j.com/developer/cla/) so that we have the right to compile and merge your code to our codebase?

Thanks again for helping us on this issue!
Zhen

liutec:
Hello @zhenlineo,

thank you, I tested the changes from #206 and the issue is gone. 

Sorry about the mess, I'm not a big contributor on GitHub and I still have _a lot_ to learn about the process (and Python). I appreciate your patience.

I signed the CLA as per the instructions.

As for ```ChunkedOutputBuffer#write``` because of ```while to_write > 0:``` and ```wrote = min(to_write, remaining)``` I think the condition ```remaining < to_write <= max_chunk_size``` may be true only on the first iteration and will cut short the existing chunk (I did not include this because I could not see any case this would be useful -- the old code cuts short the existing/last new chunk because it does ```if new_data_size > chunk_remaining: self.chunk()``` without filling ```chunk_remaining```). Also, ```self.chunk()``` without ```remaining = max_chunk_size``` will waste one iteration (3 comparisons and 2 ops)

I just noticed the junk code I've committed:
```python
data_size = self._end - self._start
if data_size > new_data_size:
    new_end = self._end + new_data_size
    self._data[self._end:new_end] = bytearray(data_size)
```
I was trying to pre-calculate the size for `_data` before the actual writes to avoid multiple extends (did not finish, hope to do so as time permits). 

From what I saw [here](https://github.com/python/cpython/blob/e2f92de6a90ae11a6d8e01bd17fd88b005516835/Objects/bytearrayobject.c#L1680), bytearrays are stored in contiguous blocks of memory and even though Python tries to [overallocate](https://github.com/python/cpython/blob/e2f92de6a90ae11a6d8e01bd17fd88b005516835/Objects/bytearrayobject.c#L227) for small changes this still implies a _huge_ overhead for the allocs & [copy](https://github.com/python/cpython/blob/e2f92de6a90ae11a6d8e01bd17fd88b005516835/Objects/bytearrayobject.c#L248) required for large data. 

Resizing ```_data``` in one shot before the chunked write starts (recursive or not) should fix the memory issue by using one extend per ```write``` call instead of one per chunk required for the data --- in the worst case. The performance gain should also be significant because the old data will remain in place.

zhenlineo:
@liutec 
I did not find your CLA, could you resent it and also cc me (zhen.li at neo4j.com) in the email?

You are most welcome to improve the fix. I am mostly trying to get in this fix first and send a patch out so that we could first clear people's problem.

Cheers,
Zhen

liutec:
@zhenlineo 
Sent it again, this time from my gmail account. Hope there will be no more problems.

zhenlineo:
@liutec Thanks very much. I got your CLA and merge the other PR with your commits.

You are most welcome to improve the code.
Thanks a again for helping us find and fix the issue.

Zhen

zhenlineo:
Hi @liutec,
 
This PR have some conflicts with the new 1.5 code as both you and me modified the same part of code. If you would like to continue working on this PR for improving buffer usage, then it might be easier to create a new PR on the top of the new 1.5 branch?

The PR #206 includes your first few commits, but I modified a bit based on your feedback. So thanks again for contribute to this driver! And if you issue new PRs, our integrated build server will recognize you as a old friend. So you are most welcome to contribute more if you like. I am closing this PR now as I am just tidying things up around before Christmas. 

Cheers,
Zhen

PR #209, opened 2017-12-19
(Do not merge) Changes to get reset to work within sessions and transactions

eastlondoner:
all tests, including integration tests, pass for me but something is up with the unpacker code - I had to silence it complaining about nothing to unpack.

There's a load of excess debug logging left in as well.


technige:
Sorry but we won't be merging this. We decided a while back not to implement reset functionality in any other drivers and will instead introduce equivalent (but more robust) functionality in a future protocol version.

PR #214, opened 2018-01-08
Added option for insecure connections from CLI

technige:


zhenlineo:
Do we on purpose hide this for internal use only? Otherwise shall we consider documenting this feature in project readme?

technige:
The CLI interface isn't mentioned in the README as it's more of an internal convenience than a recommended interface.

PR #215, opened 2018-01-09
Added bookmarks example

technige:


mariascharin:
@technige Could we have it in 1.5 in order to make it available right away, rather than waiting for 1.5? It's been put into 1.5 for the other drivers.

zhenlineo:
@mariascharin Failed to change on this one, but here is the PR for 1.5 https://github.com/neo4j/neo4j-python-driver/pull/216

PR #221, opened 2018-03-02
keep track of the hostname along with the IP information for all addr…

eastlondoner:
…esses

Basically an 'address' has to be a unique combination of an IP address, a port and a hostname.

You can have all permutations of each with any combination of fields being duplicated.


eastlondoner:
This is no longer required. Issue has been addressed in:
https://github.com/neo4j/neo4j-python-driver/pull/230/files

PR #225, opened 2018-03-18
Better type identification

funkyfuture:
This allows to pass objects that are not based on list or dict, but implement their types.

### DIFF
--- a/neo4j/v1/types/__init__.py
+++ b/neo4j/v1/types/__init__.py
@@ -26,6 +26,7 @@ into concrete values through the PackStreamHydrant.
 """
 
 
+from collections import Mapping, Sequence
 from functools import reduce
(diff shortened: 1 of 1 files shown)


technige:
This is an undesirable change as it allows types to be accepted as parameters that are not directly supported, leading to potential unexpected behaviours. `List` and `Map` are specific Cypher types and deliberately have a 1-to-1 mapping to `list` and `dict` in Python.

To pass derived types as parameters, coerce them explicitly first: `list(sequence)` or `dict(mapping)`.

Closing PR.

funkyfuture:
isn't it a user's responsibility to provide data that the driver can handle? i mean, when i pass an object to a function that expects something that behaves like a `Mapping` as `dict` does, i have to make sure that this object does so. not the driver.

PR #226, opened 2018-03-19
add neo4j.v1.types to setup.py

eastlondoner:
I think there are ways of discovering submodules programatically when setup.py is run but this fixes installing the driver with pip for now. Without it the types module does not get installed and it fails at runtime


technige:
This has been fixed in a separate commit.

PR #233, opened 2018-04-06
Base Exception Class

KyleJamesWalker:
Just create a single base exception to allow easier catching or errors.

### DIFF
--- a/neo4j/exceptions.py
+++ b/neo4j/exceptions.py
@@ -24,27 +24,32 @@ This module contains the core driver exceptions.
 """
 
 
-class AddressError(Exception):
+class Neo4jError(Exception):
+    """ Base Exception.
+    """
+
+
+class AddressError(Neo4jError):
     """ Raised when a network address is invalid.
     """
 
 
-class ProtocolError(Exception):
+class ProtocolError(Neo4jError):
     """ Raised when an unexpected or unsupported protocol event occurs.
     """
 
 
-class ServiceUnavailable(Exception):
+class ServiceUnavailable(Neo4jError):
     """ Raised when no database service is available.
     """
 
 
-class SecurityError(Exception):
+class SecurityError(Neo4jError):
     """ Raised when an action is denied due to security settings.
     """
 
 
-class CypherError(Exception):
+class CypherError(Neo4jError):
     """ Raised when the Cypher engine returns an error to the client.
     """
 


KyleJamesWalker:
I can't view the check's error, can someone let me know how the error so I can fix it?

technige:
Given that each of these errors should generally be handled by a different code path, what is the use case for this?

KyleJamesWalker:
I want to be able to catch any Neo4j error without having to catch them all. @technige 

For example right now I have the following in a flask exception handler:
``` python
    @app.errorhandler(neo4j.exceptions.ServiceUnavailable)
    @app.errorhandler(neo4j.exceptions.AddressError)
    @app.errorhandler(neo4j.exceptions.ProtocolError)
    @app.errorhandler(neo4j.exceptions.SecurityError)
    @app.errorhandler(neo4j.exceptions.CypherError)
    def handle_neo4j_errors(e):
```

technige:
I would argue that your code demonstrates an undesirable pattern of usage. Errors should be handled in a way appropriate to the condition that the error is signalling, the code that is being executed when the error occurs, and the action that the application should consequently take.

- `ServiceUnavailable` signals that the driver has been unable to communicate with the server or cluster. This could be handled by backing off and retrying later or alerting the end user to the communication failure.
- `AddressError` signals that the address of a server cannot even be resolved. This could indicate a configuration error or could request corrected input from an end user.
- `ProtocolError` happens when there is a fundamental breakdown in communication between the driver and the server. This _should_ never happen, but if it were to occur, it would indicate a bug in the driver or server software. Neither the application nor the end user can do anything about this (except perhaps work around it) and a bug report should be raised.
- `SecurityError` may indicate that the user's credentials are incorrect or have expired. Maybe a prompt to the end user for correct details is appropriate here.
- `CypherError` would occur when a Cypher statement fails. There are several reasons for this (and several appropriate subclasses) but this will often mean either a transient error (where a delay and retry is appropriate) or a fault in your application code. Unless of course the user enters Cypher directly, in which case the error should be forwarded to the user for display.

KyleJamesWalker:
In this case I want to simply present the error to a developer and then report to Sentry.  When I'm catching and handling errors to be recovered I want to handle them separately.  Quite a few libraries have a universal base exception so I can catch and report easily, while handing is done an a individual exception, or group of exceptions like you mentioned before.

technige:
Closing as wontfix.

PR #234, opened 2018-04-06
Base Exception Class

KyleJamesWalker:
Just create a single base exception to allow easier catching or errors.

Note: I also made a PR for 1.5 also. I wasn't sure what one I should do or if I should do both.


technige:
Closing as wontfix.

PR #241, opened 2018-05-16
[1.7.0] Custom resolver option

technige:
This PR adds an option to pass a custom resolver function into the driver configuration. This allows an override of the regular `getaddrinfo` call used for DNS resolution, introducing the possibility for multiple addresses to be provided for bootstrapping the routing process.

### DIFF
--- a/docs/source/driver.rst
+++ b/docs/source/driver.rst
@@ -160,7 +160,7 @@ For example::
             from socket import gaierror
             raise gaierror("Unexpected socket address %r" % socket_address)
 
-     driver = GraphDatabase.driver("bolt://foo:9999", auth=("neo4j", "password"), resolver=my_resolver)
+     driver = GraphDatabase.driver("bolt+routing://foo:9999", auth=("neo4j", "password"), resolver=my_resolver)
 
 
 
--- a/docs/source/driver.rst
+++ b/docs/source/driver.rst
@@ -147,8 +147,9 @@ This setting does not terminate running queries.
 ``resolver``
 ------------
 
-A custom resolver function to use for DNS resolution.
+A custom resolver function to resolve host and port values ahead of DNS resolution.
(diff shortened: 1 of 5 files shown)


docs/source/driver.rst
```diff
@@ -144,6 +144,24 @@ The maximum time to allow for retries to be attempted when using transaction fun
 After this time, no more retries will be attempted.
 This setting does not terminate running queries.
 
+``resolver``
+------------
+
+A custom resolver function to use for DNS resolution.
```
lutovich: Maybe we can say that it is a hostname or IP address resolver and default implementation performs DNS resolution.

eastlondoner:
I'm not sure this should replace getaddrinfo.

I'd rather that it just comes before getaddrinfo as a straight transform - so I can return new hostnames or ip addresses

e.g. in AWS I might want to transform an external hostname to an internal hostname using some regex - but I still want you to do the DNS resolution.

technige:
That's an interesting idea. Maybe we should also fall back to `getaddrinfo` not just when no resolver is supplied but also if it returns nothing.

docs/source/driver.rst
technige: :+1: Yes, that's better wording.

eastlondoner:
What I'd like is something that works with

```
def my_resolver(socket_address):
         target_host, target_port = socket_address
         my_config = json.loads("/my-config-mount/latest.json")
         results = my_config[target_host]
         for x in results:
             yield x.host, x.port
```

given

```
### latest.json
{
  "dev": [
    {
      "host": "127.0.0.1"
      "port": 7687
    }
  ],
  "production": [
    {
      "host": "andy.neo4j.io"
      "port": 7000
    }
  ]
}
```

usage example
`neo4j.driver(f"bolt+routing://dev")`

eastlondoner (approved):

PR #257, opened 2018-09-14
No STDOUT messages

danodonovan:
Could [these lines](https://github.com/neo4j/neo4j-python-driver/blob/1.7/neo4j/__init__.py#L654) be cut from `neo4j/__init__.py`? 

```
metadata = getattr(unit_of_work, "metadata", None)
print(metadata) # TODO
```
The metadata print fills my console 👎 and isn't controllable with `logging`.

### DIFF
--- a/neo4j/__init__.py
+++ b/neo4j/__init__.py
@@ -651,8 +651,6 @@ class Session(object):
         if not callable(unit_of_work):
             raise TypeError("Unit of work is not callable")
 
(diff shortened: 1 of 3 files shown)


danodonovan:
To cover  #258

danodonovan:
I'm afraid I can't see the TeamCity reasons for the failing tests - tests pass locally with `tox`. I'm happy to fix the PR if someone can share the errors!

PR #289, opened 2019-05-14
Add routing_driver -method to GraphDatabase -class

psikala:
This adds routing_driver -method to GraphDatabase -class which allows passing core -members as comma separated URI -string or as a list. Approach is similar than neo4j-java-driver implementation.

### DIFF
--- a/neo4j/__init__.py
+++ b/neo4j/__init__.py
@@ -76,8 +76,29 @@ class GraphDatabase(object):
         """ Create a :class:`.Driver` object. Calling this method provides
         identical functionality to constructing a :class:`.Driver` or
         :class:`.Driver` subclass instance directly.
+        :param uri: the URL to a Neo4j instance
+        :param config: user defined configuration
+        :return: a new driver to the database instance specified by the URL
         """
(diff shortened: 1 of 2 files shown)


technige:
Thank you for your contribution, but we do not plan to support this API in drivers other than the Java driver at this time as it encourages a non-recommended usage pattern. The Driver URI should generally be the (single) _logical_ address of a DBMS, not the _physical_ addresses of multiple cluster members, so as to avoid requiring recreation of the `Driver` object on core membership changes.

The logical-to-physical mapping can be achieved using DNS, an intermediary network component such as a load balancer, or a custom resolver function (for a pure in-application solution). Please see the manual for more details:
https://neo4j.com/docs/driver-manual/current/client-applications/#_cluster_bootstrap_address_resolution

psikala:
Ok. Thank you for the quick response. Using DNS -based -resolution or custom resolver sounds reasonable.

Reason for this was that correct usage pattern is not clearly documented/recommended. Neo4j OGM driver also supports similar URI -pattern so it's little bit of confusing.

Issue #290, opened 2019-05-15
DB connection error when shutting down the flask server


mahdizadehms:
I got DB connection error when shutting down the flask server (using CTRL+C). The error has no impact on the app execution process or the user experience (It happens only when you shut down the Flask app).

```
 * Serving Flask app "app" (lazy loading)
 * Environment: development
 * Debug mode: on
 * Running on http://127.0.0.1:5000/ (Press CTRL+C to quit)
 * Restarting with stat
 * Debugger is active!
 * Debugger PIN: 124-357-501
^C
Failed to write data to connection Address(host='127.0.0.1', port=7687) ("0; 'Underlying socket connection gone (_ssl.c:2067)'")
Failed to write data to connection Address(host='127.0.0.1', port=7687) ("0; 'Underlying socket connection gone (_ssl.c:2067)'")
```

Flask version: 1.0.2
Neo4j Server version: 3.5.4 (community)
Driver version: Python driver 1.7.2
Operating System: Ubuntu 18.04 on Windows 10 WSL  

### Steps to reproduce
1. Start the Flask app and initialize GraphDatabse driver (No need to initiate any sessions)
1. Close the Flask app using CTRL+C

### Codes
movies.py:
```
#!/usr/bin/env python
from flask import Flask, g

from neo4j import GraphDatabase, basic_auth

app = Flask(__name__)

driver = GraphDatabase.driver('bolt://localhost',auth=basic_auth("neo4j", 'a'))

def get_db():
    if not hasattr(g, 'neo4j_db'):
        g.neo4j_db = driver.session()
    return g.neo4j_db

@app.teardown_appcontext
def close_db(error):
    if hasattr(g, 'neo4j_db'):
        g.neo4j_db.close()

if __name__ == '__main__':
    app.run(port=8080)
```

requirements.txt:
```
Flask==1.0.2
neo4j==1.7.2
```

run:
`python movies.py`

ghost:
Yeah I got this too, you can get the same result just initializing the driver in a normal python script. If you read [https://neo4j.com/developer/python/](url) you will see they use `driver.close()` when cleaning up. Running that will not produce the error shown above.

ali-ince:
Since this is resolved by correct usage.

PR #297, opened 2019-08-04
(docs) use with-as for driver initialization

ComFreek:
In particular, the current example does not `close()` the connection causing two (non-critical) warnings appear at termination of the Python script: https://github.com/neo4j/neo4j-python-driver/issues/290

Using `with` exactly prevents forgotten `close()` calls.


martin-neotech:
This issue have been resolved with 4.0 releases.

PR #320, opened 2019-12-25
Added tuple as valid dehydrate type

AdallomRoy:
No reason not to be able to use tuple the same way list is used.


technige:
Hi @AdallomRoy. Thank you for your submission.

While this would certainly be possible, technically speaking, we have made a conscious choice not to add too much magic to the type mapping logic. Python lists are the correct analogue for the Cypher List type and tuples, while similar, generally present a semantic that is closer to a record than a list.

To coerce tuples to lists for use in queries, please add an explicit `list(...)` wrapper to the value.

Closing as wontfix.

Issue #335, opened 2020-01-17
How to connect to neo4j 4.0 using this driver?


naughtyGitCat:
Does driver 1.7.6 support neo4j4.0?

```bash
  File "/home/dba/db_neo4j/adapters/neo4j.py", line 24, in initialize_neo4j_pool
    password=Neo4jConfig.password))
  File "/home/dba/db_neo4j/python3/lib/python3.7/neo4j/__init__.py", line 120, in driver
    return Driver(uri, **config)
  File "/home/dba/db_neo4j/python3/lib/python3.7/neo4j/__init__.py", line 161, in __new__
    return subclass(uri, **config)
  File "/home/dba/db_neo4j/python3/lib/python3.7/neo4j/__init__.py", line 235, in __new__
    pool.release(pool.acquire())
  File "/home/dba/db_neo4j/python3/lib/python3.7/neobolt/direct.py", line 715, in acquire
    return self.acquire_direct(self.address)
  File "/home/dba/db_neo4j/python3/lib/python3.7/neobolt/direct.py", line 608, in acquire_direct
    connection = self.connector(address, error_handler=self.connection_error_handler)
  File "/home/dba/db_neo4j/python3/lib/python3.7/neo4j/__init__.py", line 232, in connector
    return connect(address, **dict(config, **kwargs))
  File "/home/dba/db_neo4j/python3/lib/python3.7/neobolt/direct.py", line 972, in connect
    raise last_error
  File "/home/dba/db_neo4j/python3/lib/python3.7/neobolt/direct.py", line 963, in connect
    s, der_encoded_server_certificate = _secure(s, host, security_plan.ssl_context, **config)
  File "/home/dba/db_neo4j/python3/lib/python3.7/neobolt/direct.py", line 854, in _secure
    s = ssl_context.wrap_socket(s, server_hostname=host if HAS_SNI and host else None)
  File "/home/dba/db_neo4j/python3/lib/python3.7/ssl.py", line 412, in wrap_socket
    session=session
  File "/home/dba/db_neo4j/python3/lib/python3.7/ssl.py", line 850, in _create
    self.do_handshake()
  File "/home/dba/db_neo4j/python3/lib/python3.7/ssl.py", line 1108, in do_handshake
    self._sslobj.do_handshake()
OSError: [Errno 0] Error
```

naughtyGitCat:
```bash
Collecting neo4j==4.0
[...]

PR #336, opened 2020-01-17
Improve dist functions

krisgeus:
#335


technige:
These scripts are used by our internal release process and, as such, we can't accept external PRs on them.

PR #339, opened 2020-01-20
Adds Bolt 4.0 protocol support at connector level

technige:


technige:
TODO: find out whether `qid` is optional before merging

PR #393, opened 2020-05-20
Make tests green when running against 4.1 db

2hdddg:
4.1 uses the correctly classified error when client aborts transaction due to timeout, need to handle that in timeout tests.

4.1 suffixes plan operators with db name, handle this and unsuffixed cases in tests.


martin-neotech (approved):
Looks nice

PR #475, opened 2020-11-13
Proper initialization of system CAs on Windows

2hdddg:
Should use SSLContext.load_default_certs to load system CAs properly on
all platforms including Windows. SSLContext.set_default_verify_paths
worked fine on Linux based OSes. load_default_certs is new in Python
3.4.

### DIFF
--- a/neo4j/conf.py
+++ b/neo4j/conf.py
@@ -261,7 +261,10 @@ class PoolConfig(Config):
             ssl_context.check_hostname = False
             ssl_context.verify_mode = ssl.CERT_NONE     # https://docs.python.org/3.5/library/ssl.html#ssl.CERT_NONE
 
-        ssl_context.set_default_verify_paths()  # https://docs.python.org/3.5/library/ssl.html#ssl.SSLContext.set_default_verify_paths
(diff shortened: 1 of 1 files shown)


neo4j/conf.py
```diff
@@ -261,7 +261,10 @@ def get_ssl_context(self):
             ssl_context.check_hostname = False
             ssl_context.verify_mode = ssl.CERT_NONE     # https://docs.python.org/3.5/library/ssl.html#ssl.CERT_NONE
 
-        ssl_context.set_default_verify_paths()  # https://docs.python.org/3.5/library/ssl.html#ssl.SSLContext.set_default_verify_paths
+        # Must be load_default_certs, not set_default_verify_paths to work
+        # on Windows with system CAs.
```
bigmontz, line 265: I think it worths mention the docs (https://docs.python.org/3.5/library/ssl.html#ssl.SSLContext.load_default_certs)

2hdddg: Hmm, I was a bit annoyed over those references all over the code ;-) I mean finding info on standard python library functions isn't that hard to google

PR #479, opened 2020-11-26
Initial testkit support

2hdddg:
Datatypes tests passing in testkit.

### DIFF
--- a/testkitbackend/backend.py
+++ b/testkitbackend/backend.py
@@ -1,0 +1,78 @@
+# Copyright (c) 2002-2020 "Neo4j,"
+# Neo4j Sweden AB [http://neo4j.com]
+#
+# This file is part of Neo4j.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


testkitbackend/backend.py
```diff
@@ -0,0 +1,63 @@
+from json import loads, dumps
+from inspect import getmembers, isfunction
+import testkitbackend.requests as requests
+
+
+# Collect all request handlers
+requestHandlers = dict([m for m in getmembers(requests, isfunction)])
```
bigmontz: I think it could be a good idea or inject this information on the constructor as param (so, the Server will create the object and pass to Backend constructor) or it be created on the Backend constructor (which is a bit less performatic). Global variables could have strange behaviours if some one change it value by 'accident'

bigmontz (approved):
🦖

testkitbackend/backend.py
2hdddg: I move it into the backend, good point!

PR #483, opened 2020-12-09
Updated ServerInfo class with new methods

technige:
This PR updates the `ServerInfo` class to add new accessors for `address`, `protocol_version` and `connection_id`. It also deprecates the old `version_info` method (for removal in 5.0) and improves the implementation of the "supported server product" check.

### DIFF
--- a/neo4j/api.py
+++ b/neo4j/api.py
@@ -191,7 +191,9 @@ class ServerInfo:
     @property
     def protocol_version(self):
         """ Bolt protocol version with which the remote server
-        communicates.
+        communicates. This is returned as a :class:`.Version`
+        object, which itself extends a simple 2-tuple of
+        (major, minor) integers.
         """
         return self._protocol_version
 
--- a/neo4j/api.py
+++ b/neo4j/api.py
@@ -26,6 +26,8 @@ from.exceptions import (
     DriverError,
     ConfigurationError,
 )
+from .meta import deprecated
+
 
 """ Base classes and helpers.
 """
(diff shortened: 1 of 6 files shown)


neo4j/api.py
```diff
@@ -172,23 +174,44 @@ def values(self):
...
+
+    @property
+    def address(self):
+        """ Network address of the remote server.
+        """
+        return self._address
+
+    @property
+    def protocol_version(self):
+        """ Bolt protocol version with which the remote server
```
2hdddg, line 193: Perhaps indicate format of version. Tuple of major, minor or string?

neo4j/api.py
```diff
@@ -172,23 +174,44 @@ def values(self):
...
         """
-        # Example "Neo4j/4.0.5"
-        # Example "Neo4j/4"
-        return self.metadata.get("server")
+        return self._metadata.get("connection_id")
 
+    # TODO in 5.0: remove this method
+    @deprecated("The version_info method is deprecated, please use "
+                "ServerInfo.agent, ServerInfo.protocol_version, or "
+                "call the dbms.components procedure instead")
```
2hdddg, line 216: Isn't dbms.components always the appropriate action?

neo4j/api.py
```diff
@@ -220,14 +243,12 @@ def version_info(self):
                 pass
         return tuple(value)
 
-    def _update_metadata(self, metadata):
-        """Internal, update the metadata and perform check that the prefix is whitelisted by calling self.version()
-
-        :param metadata: metadata from the server
-        :type metadata: dict
+    def update(self, metadata):
```
2hdddg, line 248: Why not internal anymore?

neo4j/io/_bolt3.py
```diff
@@ -153,9 +155,10 @@ def hello(self):
             logged_headers["credentials"] = "*******"
         log.debug("[#%04X]  C: HELLO %r", self.local_port, logged_headers)
         self._append(b"\x01", (headers,),
-                     response=InitResponse(self, on_success=self.server_info._update_metadata))
+                     response=InitResponse(self, on_success=self.server_info.update))
         self.send_all()
         self.fetch_all()
+        check_supported_server_product(self.server_info.agent)
```
2hdddg, line 161: This would raise if agent string changes. Why not just treat the agent string as whatever string?

neo4j/api.py
technige: It depends. Due to how the version information has been collected historically, there may be several reasons why users are calling this method. Drawing attention to all methods can only help.

neo4j/api.py
technige: This was being called from separate modules (Bolt 3 and Bolt 4 handler modules). Such trans-module calls are not considered internal and this was therefore unpythonic. The IDE highlights it as a bad code smell.

2hdddg (approved):

PR #488, opened 2020-12-21
Do not connect to cluster/server upon driver creation

2hdddg:
Since driver is typically created at application startup a failure to
connect to a server or cluster that is temporarly down would cause the
application to not start either. Clients should use verify connection to
check for proper configuration.

### DIFF
--- a/testkitbackend/requests.py
+++ b/testkitbackend/requests.py
@@ -98,6 +98,13 @@ def SessionBeginTransaction(backend, data):
     backend.send_response("Transaction", {"id": key})
 
 
+def SessionLastBookmarks(backend, data):
+    key = data["sessionId"]
+    session = backend.sessions[key].session
+    bookmark = session.last_bookmark()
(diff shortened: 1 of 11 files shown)


testkitbackend/requests.py
```diff
@@ -90,19 +93,80 @@ def SessionBeginTransaction(backend, data):
...
     if timeout:
-        timeout = int(timeout)
+        timeout = float(timeout) / 1000
     tx = session.begin_transaction(metadata=metadata, timeout=timeout)
     key = backend.next_key()
     backend.transactions[key] = tx
     backend.send_response("Transaction", {"id": key})
 
 
+def SessionReadTransaction(backend, data):
```
technige, line 103: This is non-standard casing for Python functions.

2hdddg: It's a bit of laziness... the requests in testkit are mapped directly to function names, didn't have the energy to change casing of the requests

vEpiphyte:
Can this PR be added to the change log for v4.3.x ?

PR #497, opened 2021-02-15
Fix spatial types allowed for non-float coordinates

### LINKED ISSUE #496: Crash when using spatial type as parameters
Hi folks!

I can't seem to get the spatial type working, just getting a crash (stack trace attached).

- Neo4j version: Enterprise 4.2.1
- Neo4j Mode: Single instance
- Python 3.9
- Driver version:  driver 4.2.1
- Operating system: Ubuntu 20.04

- **Steps to reproduce**: run the following piece of code:

```
import neo4j
print(neo4j.__version__)

from neo4j.spatial import WGS84Point
from neo4j import GraphDatabase, basic_auth
driver = GraphDatabase.driver("bolt://localhost:7687", auth=basic_auth("neo4j", ""), encrypted=False)
point = WGS84Point(dict(x=2, y=48))
cypher = "WITH $point as point RETURN [point.x, point.y]"
params = {"point": point}
with driver.session() as session:
    session.run(cypher, params)
```

- Expected behavior : query returns [2, 48] (or a meaningful error message in case of misuse)
- Actual behavior: `neo4j.exceptions.ServiceUnavailable: Failed to read from defunct connection IPv4Address(('localhost', 7687)) (IPv4Address(('127.0.0.1', 7687)))`

Full stack trace is attached.
[neo_spatial.log](https://github.com/neo4j/neo4j-python-driver/files/5968103/neo_spatial.log)

Also, only error in server logs is:

```
2021-02-11 18:00:06.353+0000 ERROR Client triggered an unexpected error [Neo.DatabaseError.General.UnknownError]: Unknown statement ID: -1. Existing IDs: [], reference 1c40d17c-1c03-4b40-b422-ceeef7fc49ec.
```

- Additional information:
    -  Same error with CartesianPoint  
    - Doing similar things with DateTime works well (and basically, anything else is working just fine :) )
    - I havn't found an example using spatial types in the doc, but the example is taken from https://github.com/neo4j/neo4j-python-driver/blob/4.3/tests/integration/test_spatial_types.py (I think)
    - Replacing the query with a `MATCH (n:Node) RETURN n.point` returns a `neo4j.spatial.WGS84Point` as expected
    - Same error as https://github.com/neo4j/neo4j-python-driver/issues/484 but, not sure the root cause is the same though.

robsdedude:
This improves the handling of arguments to spatial types that are not floats.

Closes #496

### DIFF
--- a/neo4j/spatial/__init__.py
+++ b/neo4j/spatial/__init__.py
@@ -53,7 +53,7 @@ class Point(tuple):
     srid = None
 
     def __new__(cls, iterable):
-        return tuple.__new__(cls, iterable)
+        return tuple.__new__(cls, map(float, iterable))
 
     def __repr__(self):
         return "POINT(%s)" % " ".join(map(str, self))
--- a/tests/unit/spatial/test_cartesian_point.py
+++ b/tests/unit/spatial/test_cartesian_point.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+# coding: utf-8
+
+# Copyright (c) 2002-2020 "Neo4j,"
+# Neo4j Sweden AB [http://neo4j.com]
+#
+# This file is part of Neo4j.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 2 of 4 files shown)


robsdedude:
reopened as #498 for CI testing purposes

PR #516, opened 2021-03-25
Don't try to close already broken connection during HELLO

robsdedude:
Previously, if the driver lost connection while sending the `HELLO` message or waiting for the response, the driver would try to send a `GOODBYE` message over the broken wire.


neo4j/io/__init__.py
```diff
@@ -258,6 +258,8 @@ def open(cls, address, *, auth=None, timeout=None, routing_context=None, **pool_
 
         try:
             connection.hello()
+        except (SessionExpired, ServiceUnavailable):
+            raise
```
technige: Instead of this extra block in the outer try, I suggest we might want to create an internal method to handle log.debug, shutdown and close, which silences all I/O Errors if they occur.

This assumes that _no_ cases of socket closure would ever want to surface such an error.

PR #517, opened 2021-03-25
Don't retry a failed commit

robsdedude:
Session would retry a transaction even if it failed on commit. This is not desirable because the driver can't know if the transaction got successfully committed on the server's side or not.


robsdedude:
Actually there is a `BoltIncompleteCommitError` that's meant for internal use only. That's why Transaction translates it to a `ServiceUnavailable` Error.

PR #518, opened 2021-03-25
Discard transaction on disconnect

robsdedude:
A transaction is tied to the connection. If the connection dies, the transaction should die too. Previously, the session would try to rollback a transaction even if the transaction's connection died earlier.

To implement this, I decided to add an `on_network_error` callback to `Result` and `Transaction`. While at it, I also refactored all `IOError`s to `OSError`s. The latter one is the canonical form for this type of error since Python 3.3 (https://docs.python.org/3/whatsnew/3.3.html#pep-3151-reworking-the-os-and-io-exception-hierarchy)


neo4j/io/_bolt4.py
```diff
@@ -389,11 +376,20 @@ def fetch_message(self):
 
         return len(details), 1
 
-    def _set_defunct(self, error=None):
-        direct_driver = isinstance(self.pool, BoltPool)
+    def _set_defunct_read(self, error=None):
+        message = "Failed to read from defunct connection {!r} ({!r})".format(
```
technige, line 384: Best to correct this to use lazy interpolation through the logging framework, e.g. `log.info("Thing %r failed", thing)`.

neo4j/work/result.py
```diff
@@ -20,20 +20,50 @@
...
+from neo4j._exceptions import BoltIncompleteCommitError
 from neo4j.data import DataDehydrator
+from neo4j.exceptions import (
+    ServiceUnavailable,
+    SessionExpired,
+)
 from neo4j.work.summary import ResultSummary
 
 
+class _ConnectionErrorHandler:
```
robsdedude, line 33: Add docstring. What does this $thing do?

neo4j/work/result.py
```diff
@@ -97,7 +127,7 @@ def on_failed_attach(metadata):
             on_failure=on_failed_attach,
         )
         self._pull()
-        self._connection.send_all()
+        self._connection._send_all()
```
robsdedude: Why do I need to call an internal method?

neo4j/work/simple.py
```diff
@@ -243,10 +254,20 @@ def _transaction_closed_handler(self):
...
+
+    def _open_transaction(self, *, access_mode, database, metadata=None,
+                          timeout=None):
         self._connect(access_mode=access_mode, database=database)
-        self._transaction = Transaction(self._connection, self._config.fetch_size, self._transaction_closed_handler)
-        self._transaction._begin(database, self._bookmarks, access_mode, metadata, timeout)
+        self._transaction = Transaction(
+            self._connection, self._config.fetch_size,
+            self._transaction_closed_handler,
+            self._transaction_network_error_handler)
```
robsdedude: Move `)` to next line.

neo4j/io/_bolt4.py
robsdedude: Actually, since this message is also used as argument to the error that `_set_defunct` raises, it will need to be interpolated sooner or later anyway.

PR #528, opened 2021-04-07
Testkit compliance

robsdedude:
Adding and adjusting testkit back end to
 * Skip tests.
 * Reduce the time the driver attempts retries.
   Testkit will cut the connection to the driver back end if the driver takes too long (10 seconds) to respond. In a controlled environment like testkit, there shouldn't be a need for very long retry periods anyway.
 * Send a traceback with errors so it's easier to debug failed tests.

### DIFF
--- a/testkitbackend/requests.py
+++ b/testkitbackend/requests.py
@@ -133,9 +133,7 @@ def NewSession(backend, data):
             "default_access_mode": access_mode,
             "bookmarks": data["bookmarks"],
             "database": data["database"],
-            "fetch_size": data.get("fetchSize", None),
-            # testkit breaks up connection after 10 seconds
-            "max_transaction_retry_time": 5,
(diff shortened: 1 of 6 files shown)


testkitbackend/requests.py
```diff
@@ -133,7 +133,9 @@ def NewSession(backend, data):
             "default_access_mode": access_mode,
             "bookmarks": data["bookmarks"],
             "database": data["database"],
-            "fetch_size": data.get("fetchSize", None)
+            "fetch_size": data.get("fetchSize", None),
+            # testkit breaks up connection after 10 seconds
+            "max_transaction_retry_time": 5,
```
robsdedude: Find a proper solution

robsdedude: Fixed it in testkit

PR #529, opened 2021-04-07
Try all routers before giving up

robsdedude:
When attempting to fetch a routing table, the driver shouldn't give up if the first attempt fails but other routers are known.

Remark: we could still do better, but this would require a huge rewrite and might introduce breaking changes.
If a custom resolver function return multiple addresses, the driver will only attempt to fetch a RT from the first server that yields a successful handshake. If that server fails to provide a valid RT for whatever reason, the driver will move on to the next known router. The other addresses returned by the custom resolver function will be ignored.

### DIFF
--- a/neo4j/io/__init__.py
+++ b/neo4j/io/__init__.py
@@ -754,10 +754,13 @@ class Neo4jPool(IOPool):
         """
         log.debug("Attempting to update routing table from {}".format(", ".join(map(repr, routers))))
         for router in routers:
-            new_routing_table = self.fetch_routing_table(
-                address=router, timeout=self.pool_config.connection_timeout,
-                database=database, bookmarks=bookmarks
-            )
+            try:
+                new_routing_table = self.fetch_routing_table(
+                    address=router, timeout=self.pool_config.connection_timeout,
+                    database=database, bookmarks=bookmarks
+                )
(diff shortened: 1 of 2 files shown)


neo4j/io/__init__.py
```diff
@@ -754,10 +754,13 @@ def update_routing_table_from(self, *routers, database=None,
...
         for router in routers:
-            new_routing_table = self.fetch_routing_table(
-                address=router, timeout=self.pool_config.connection_timeout,
-                database=database, bookmarks=bookmarks
-            )
+            try:
+                new_routing_table = self.fetch_routing_table(
+                    address=router, timeout=self.pool_config.connection_timeout,
+                    database=database, bookmarks=bookmarks
+                )
```
robsdedude, line 761: Make this method always return `None` instead of raising an exception if anything goes wrong.

neo4j/io/__init__.py
```diff
@@ -786,8 +789,12 @@ def update_routing_table(self, *, database, bookmarks):
             ):
                 # Why is only the first initial routing address used?
```
robsdedude, line 790: pop initial router if in list and remove list comprehension further down.

robsdedude:
Closed in favor of #532

PR #535, opened 2021-04-20
Use twine for release process

robsdedude:


bin/release
```diff
@@ -10,5 +11,7 @@ then
     exit 1
 else
     source "${ROOT}/bin/dist-functions"
-    setup "${VERSION}" sdist upload -r pypi --show-response
+    twine upload --repository testpypi \
```
technige: The release script here contains the test repository.

Maybe this should be true unless the user passes in a `--real` option or something, whereupon it switches to the real repository.

PR #537, opened 2021-04-26
Align verify_connectivity with other drivers and enable testkit tests

robsdedude:
Other drivers (e.g., Java) fetch a new routing table on `verify_connectivity`. Testkit (ab)uses this behavior in tests. To make the Python driver compliant with theses tests, I adjusted it to also fetch a routing table with all usual "try all routers" and "use the initial router if no writers are known to you" logic.

Furthermore, this PR enables all tests. Some were fixed with previous PRs but just never enabled, others required slight adjustments in testkit: https://github.com/neo4j-drivers/testkit/pull/91  <- This needs to me merged first.


robsdedude:
`verify_connectivity` needs more thought and a proper mechanism for testkit to tell the driver to fetch a new RT must be found.

PR #540, opened 2021-04-30
Handle Neo.ClientError.Security.AuthorizationExpired

robsdedude:
- Close the connection and mark all idle connections in the pool as expired
 - Raise the error to the user or do a retry if inside transaction function

Refactoring shared code of Bolt 3 and Bolt 4 connection into base class.

### DIFF
--- a/neo4j/io/__init__.py
+++ b/neo4j/io/__init__.py
@@ -177,11 +177,18 @@ class Bolt(abc.ABC):
     @property
     @abc.abstractmethod
     def supports_multiple_results(self):
+        """ Boolean flag to indicate if the connection version supports multiple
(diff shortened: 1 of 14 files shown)


neo4j/io/__init__.py
```diff
@@ -107,6 +121,69 @@ class Bolt:
...
+            credentials = self.auth_dict["credentials"]
+        except KeyError:
+            pass
+        else:
+            if credentials is None:
+                raise AuthError("Password cannot be None")
+
+    @property
+    @abc.abstractmethod
+    def supports_multiple_results(self):
```
robsdedude, line 179: Add doc string

PR #554, opened 2021-06-21
Empty query should be sent to server (Query and str type)

robsdedude:


robsdedude:
This has been postponed to probably 6.0 as we can't do this as long as we want to keep (basic) support for 3.5 servers.

PR #566, opened 2021-07-20
.run on a failed transaction will also fail

robsdedude:
Mark a transaction as broken on `.run` failure. 

If executing a query fails (for whatever reason), the server will terminate the
transaction. So we might as well mark the transaction as broken on the driver
side and fail fast (before contacting the server), should the client code try to
use the transaction again.


bigmontz (approved):
🏅

PR #572, opened 2021-08-03
Don't send RESET on READY (clean) connections

robsdedude:


neo4j/io/_bolt4.py
```diff
@@ -162,24 +177,25 @@ def run(self, query, parameters=None, mode=None, bookmarks=None, metadata=None,
         fields = (query, parameters, extra)
         log.debug("[#%04X]  C: RUN %s", self.local_port, " ".join(map(repr, fields)))
         if query.upper() == u"COMMIT":
-            self._append(b"\x10", fields, CommitResponse(self, **handlers))
+            self._append(b"\x10", fields, CommitResponse(self, "run",
+                                                         **handlers))
         else:
-            self._append(b"\x10", fields, Response(self, **handlers))
+            self._append(b"\x10", fields, Response(self, "run", **handlers))
         self._is_reset = False
```
bigmontz: Is it a leftover ?

neo4j/io/_bolt4.py
```diff
@@ -225,11 +241,22 @@ def fail(metadata):
             raise BoltProtocolError("RESET failed %r" % metadata, self.unresolved_address)
 
         log.debug("[#%04X]  C: RESET", self.local_port)
-        self._append(b"\x0F", response=Response(self, on_failure=fail))
+        self._append(b"\x0F", response=Response(self, "reset", on_failure=fail))
         self.send_all()
         self.fetch_all()
         self._is_reset = True
 
+    def _update_server_state_on_success(self, metadata, message):
```
bigmontz: This method seems to appear twice with the same implementation. 

Maybe it could be the case of having a function call `get_next_state(current_state, message)` to replace the following code.

```
STATE_TRANSITIONS\
            .get(self._server_state, {})\
            .get(message, self._server_state)
```

Or maybe it could be something like `self._server_state = self._server_state.apply(message)`

bigmontz (approved):
⛑️

neo4j/io/_bolt4.py
robsdedude: Well spotted!

robsdedude:
Waiting for https://github.com/neo4j-drivers/testkit/pull/195 to re-run the TestKit suite.

PR #573, opened 2021-08-04
Close socket without GOODBYE if RESET fails.

robsdedude:
This is considered a protocol violation.


robsdedude:
This fix still sends `RESET` before closing the connection.

PR #585, opened 2021-09-14
Only send qid with PULL and DISCARD when necessary

robsdedude:
Waiting for neo4j-drivers/testkit#249 to me merged first (re-run TestKit pipeline afterwards)

### DIFF
--- a/neo4j/io/__init__.py
+++ b/neo4j/io/__init__.py
@@ -137,6 +137,11 @@ class Bolt(abc.ABC):
     #: The pool of which this connection is a member
     pool = None
 
+    # Store the id of the most recent ran query to be able reduce sent bits by
+    # using the default (-1) to refer to the most recent query when pulling
+    # results for it.
+    most_recent_qid = None
+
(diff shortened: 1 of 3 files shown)


neo4j/io/__init__.py
```diff
@@ -137,6 +137,11 @@ class Bolt(abc.ABC):
     #: The pool of which this connection is a member
     pool = None
 
+    # Store the id of the most recent ran query to be able reduce sent bits by
```
fbiville:
```suggestion
    # Store the id of the most recent ran query to be able to reduce sent bits by
```

neo4j/work/result.py
```diff
@@ -117,7 +124,10 @@ def _run(self, query, parameters, db, access_mode, bookmarks, **kwparameters):
 
         def on_attached(metadata):
             self._metadata.update(metadata)
-            self._qid = metadata.get("qid", -1)  # For auto-commit there is no qid and Bolt 3 do not support qid
+            # For auto-commit there is no qid and Bolt 3 do not support qid
```
fbiville:
```suggestion
            # For auto-commit there is no qid and Bolt 3 does not support qid
```

fbiville (approved):

PR #596, opened 2021-09-30
Rework the whole datetime library documentation

robsdedude:


neo4j/time/__init__.py
```diff
@@ -322,15 +331,73 @@ def utc_time(self):
...
+    A :class:`.Duration` represents the difference between two points in time.
+    Duration objects store a composite value of `months`, `days`, `seconds`,
+    and `nanoseconds`. Unlike :class:`datetime.timedelta` however, days, and
+    seconds/nanoseconds are never interchanged. All values except seconds and
+    nanoseconds are applied separately in calculations.
+
+    A :class:`.Duration` stores four primary instance attributes internally:
+    `months`, `days`, `seconds` and `nanoseconds`. These are maintained as
+    individual values and are immutable. Each of these four attributes can carry
+    its own sign, with the exception of `nanoseconds`, always has the smae sign
```
fbiville:
```suggestion
    its own sign, with the exception of `nanoseconds`, always has the same sign
```

neo4j/time/__init__.py
```diff
@@ -322,15 +331,73 @@ def utc_time(self):
...
+
+    The primary instance attributes and their permitted ranges are listed below.
+
+    ===============  ========================================================
+    Attribute        Value
+    ---------------  --------------------------------------------------------
+    ``months``       Between -(2\ :sup:`63`) and (2\ :sup:`63` - 1) inclusive
+    ``days``         Between -(2\ :sup:`63`) and (2\ :sup:`63` - 1) inclusive
+    ``seconds``      Between -(2\ :sup:`63`) and (2\ :sup:`63` - 1) inclusive
+    ``nanoseconds``  Between -0.999,999,999 and +0.999,999,999 inclusive
```
fbiville: Hmmm, I thought nanoseconds were always >= 0, as I understood it represented the elapsed nanoseconds (as you did with `neo4j.time.Time`).

Here's what I documented for another project:
> • -9_223_372_036_854_775_808 ≤ as_seconds(months) + as_seconds(days) +
seconds ≤ 9_223_372_036_854_775_807
• 0 ≤ nanoseconds ≤ 999_999_999

robsdedude: Go is indeed right here https://github.com/neo4j/neo4j/blob/3c961a374c84109861f190bc8cda9e949d0bb0bd/community/values/src/main/java/org/neo4j/values/storable/DurationValue.java#L76..L77
The joy of datetime libraries :scream: 
This actually means I need to change the code as well.

andres-fr:
Edit: Sorry, my comment was related to `py2neo` and not to `neo4j`. I removed it.

PR #614, opened 2021-10-28
Improve Aura Example

bigmontz:
Add examples for reading relationships and logging.

### DIFF
--- a/tests/integration/examples/test_driver_introduction_example.py
+++ b/tests/integration/examples/test_driver_introduction_example.py
@@ -25,9 +25,10 @@ from contextlib import redirect_stdout
 from io import StringIO
 
 # tag::driver-introduction-example-import[]
-from neo4j import GraphDatabase
 import logging
 import sys
+
+from neo4j import GraphDatabase
 from neo4j.exceptions import ServiceUnavailable
 # end::driver-introduction-example-import[]
 
@@ -40,7 +41,6 @@ from neo4j._exceptions import BoltHandshakeError
 class App:
 
     def __init__(self, uri, user, password):
-
         self.driver = GraphDatabase.driver(uri, auth=(user, password))
 
     def close(self):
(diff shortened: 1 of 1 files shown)


tests/integration/examples/test_driver_introduction_example.py
```diff
@@ -27,6 +27,7 @@
 # tag::driver-introduction-example-import[]
 from neo4j import GraphDatabase
 import logging
+import sys
 from neo4j.exceptions import ServiceUnavailable
```
robsdedude:
```suggestion
import logging
import sys

from neo4j import GraphDatabase
from neo4j.exceptions import ServiceUnavailable
```

First import sys modules, then 3rd party modules, then local modules. Each block in alphabetical order and separated by a blank line.

tests/integration/examples/test_driver_introduction_example.py
```diff
@@ -39,33 +40,50 @@
 class App:
 
     def __init__(self, uri, user, password):
+
```
robsdedude:
```suggestion
```

tests/integration/examples/test_driver_introduction_example.py
```diff
@@ -105,8 +124,9 @@ def test_driver_introduction_example(uri, auth):
     try:
         s = StringIO()
         with redirect_stdout(s):
+            App.enable_log(logging.INFO, sys.stdout)
```
robsdedude, line 127:
```suggestion
            App.enable_log(logging.DEBUG, sys.stdout)
```

I'm unsure if this should be `DEBUG` or `INFO`. Under normal circumstances, there is not much output (if any at all) to be observed on info level. So for illustrative purposes, debug might be the level of choice. On the flip side, I don't want to encourage people to enable debug logging just for the fun of it. It's noisy and might even be misleading if you don't know what you're looking at.

bigmontz: I was also in doubt about the level, I'm setting it as INFO since the DEBUG could be really overwhelming.

PR #630, opened 2021-12-17
Don't close stale connections while in use

robsdedude:
The pool will close connections marked as stale when trying to pick them up
from the pool. It should not do so while the connection is in use (i.e.,
already borrowed from the pool). In concurrent systems, this would lead to all
sorts of errors caused by the connection being managed from different threads:
 * the thread trying to acquire a connection will close the stale connection on
   encounter
 * while the thread that borrowed it might still be using it, e.g., to fetch
   results or run a query

### DIFF
--- a/neo4j/io/__init__.py
+++ b/neo4j/io/__init__.py
@@ -39,6 +39,7 @@ from collections import (
     defaultdict,
     deque,
 )
+import logging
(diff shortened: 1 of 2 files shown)


bigmontz (approved):
🐍

PR #663, opened 2022-03-01
Add Result.to_df to export records as pandas DataFrame

robsdedude:
To be able to add node and relationship expansion, we need 
 * ~https://github.com/neo4j/neo4j-python-driver/pull/671~
 * https://github.com/neo4j/neo4j-python-driver/pull/672

to be merged first.


bigmontz (approved):
🐼

PR #683, opened 2022-03-18
Fix `neo4j.time`

robsdedude:
* Similarly to https://github.com/neo4j/neo4j-python-driver/pull/616,
   `DateTime` objects with fixed negative UTC offset were incorrectly
   serialized before sent to the database.
 * Comparison operators of several temporal types was incorrect.


AndyHeap-NeoTech (approved):
All looks sensible to me.

PR #704, opened 2022-04-05
Remove usage of memoryview calls

robsdedude:
On paper, memoryviews seem like a good tool to increase the driver's efficiency.
In practice, they seem to yield no measurable performance improvement (probably
due to the overhead of their creation and lifetime handling), but instead cause
issues if not properly used.

:warning: DO NOT MERGE :warning: 
This PR is only meant to run the pipelines


robsdedude:
This PR has served its purpose. :wave:

PR #707, opened 2022-04-06
TestKit: add unified error class mappings

robsdedude:
Depends on https://github.com/neo4j-drivers/testkit/pull/445


robsdedude:
I does not look like this is going to happen anytime soon.

PR #724, opened 2022-05-17
Cache socket port number in bolt class

robsdedude:
The port number is (at least) queried for each debug log line. No matter if logging is enabled or not. Wrapping `socket.getsockname()` into a check whether debug logging is enabled or not, to only query it when necessary, made the driver slower for environments with a fast response (read: my local machine). Therefore, caching the port number is the only viable compromise. The downside is that this assumes that the port number never changes. As the driver is right now, it hands over all control over the socket to the Bolt class, so this assumption is safe. But it introduces complexity in the code base by adding the need to know about and adhere to this principle in future changes.

Background: a customer complained about the calls slowing down their deployment significantly. On my local machine it's actually really fast and commenting out the line in question (https://github.com/neo4j/neo4j-python-driver/blob/1ed96f94a7a59f49c473dadbb81715dc9651db98/neo4j/_async/io/_bolt5.py#L271 + in the other bolt versions) and timing a workload that hits that line frequently, didn't alter the driver's throughput in a measurable way. Timing this call in a Windows VM on my Linux machine showed it was 4 times slower. I didn't go as far as timing the whole driver in the VM.


neo4j/_async/io/_bolt.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, unresolved_address, sock, max_connection_lifetime, *,
                  auth=None, user_agent=None, routing_context=None):
         self.unresolved_address = unresolved_address
         self.socket = sock
+        self.local_port = self.socket.getsockname()[1]
+        assert self.local_port  # TODO: remove, just for testing
```
robsdedude: Remove this, before merging!

neo4j/_async/io/_common.py
```diff
@@ -42,6 +42,8 @@ class AsyncMessageInbox:
 
     def __init__(self, s, on_error):
         self.on_error = on_error
+        self._local_port = s.getsockname()[1]
+        assert self._local_port  # TODO: remove, just for testing
```
robsdedude: :eyes:

neo4j/_sync/io/_bolt.py
```diff
@@ -100,6 +100,8 @@ def __init__(self, unresolved_address, sock, max_connection_lifetime, *,
                  auth=None, user_agent=None, routing_context=None):
         self.unresolved_address = unresolved_address
         self.socket = sock
+        self.local_port = self.socket.getsockname()[1]
+        assert self.local_port  # TODO: remove, just for testing
```
robsdedude: :eyes:

AndyHeap-NeoTech:
The change is simple enough and looks ok to me. I'm wondering what the error would be if the port did happen to change without the cache knowing.

robsdedude:
> The change is simple enough and looks ok to me. I'm wondering what the error would be if the port did happen to change without the cache knowing.

Currently, the port number is only used for logging. So the worst thing that could happen are wrong and maybe misleading log messages or debug sessions.

PR #728, opened 2022-05-25
Improve logging helper `neo4j.debug`

robsdedude:
- no ANSI color codes by default
 - log thread name and id
 - add documentation for logging


bigmontz (approved):
🎨

PR #732, opened 2022-05-31
Fix pool connection ownership

robsdedude:
Fixing and testing two faulty corner-cases
 * When a connection fails, all idle connections to that host will be closed.
   If their closure fails, the same action is invoked recursively while being
   in the process of closing idle connections already. This leads to
   inconsistent data.
 * [4.4 only] When opening of a new connection fails, the pool used to
   forcefully close all connections to the host, including lent
   connections. Opening the doors wide for race conditions because connections
   are not thread safe.


AndyHeap-NeoTech (approved):
Maybe just expand the PR with explanation of the bug it's fixing.

PR #672, opened 2022-03-07
Implement bolt 5.0 + new style element ids

robsdedude:
Waiting for
* https://github.com/neo4j-drivers/testkit/pull/384
* https://github.com/neo4j-drivers/testkit/pull/423
* https://github.com/neo4j-drivers/testkit/pull/427

Merged in https://github.com/neo4j/neo4j-python-driver/pull/671 as both changes go hand in hand. Now, this PR also depends on:
 * https://github.com/neo4j-drivers/testkit/pull/408

PR #737, opened 2022-06-01
Remove null handling for Node/Relationship Ids

thelonelyvulpes:
Remove code for handling null ids inline with updated specification regarding node/ relationship ids

Amends https://github.com/neo4j/neo4j-python-driver/pull/672


robsdedude:
The driver still assumes that with freki disabled `str(id) == element_id`, which is not the case and needs fixing.

robsdedude (approved):

PR #743, opened 2022-06-03
Re-introduce session config options to `verify_connectivity`

robsdedude:
Also changes `get_server_info` that was introduced in 5.0 to accept
session config options.

Amends https://github.com/neo4j/neo4j-python-driver/pull/654

### DIFF
--- a/neo4j/_async/driver.py
+++ b/neo4j/_async/driver.py
@@ -16,6 +16,8 @@
 # limitations under the License.
 
 
+import warnings
+
 from .._async_compat.util import AsyncUtil
 from .._conf import (
     TrustAll,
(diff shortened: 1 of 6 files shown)


neo4j/_async/driver.py
```diff
@@ -347,7 +371,14 @@ async def get_server_info(self):
...
         """
-        async with self.session() as session:
+        if config:
+            experimental_warn(
+                "All configuration key-word arguments to "
+                "verify_connectivity() are experimental. They might be "
+                "changed or removed in any future version without prior "
+                "notice."
+            )
+        async with self.session(**config) as session:
```
bigmontz, line 381: We are missing tests for checking if the experimental warning is also triggered in the `get_server_info`. The PR description can also mentioned the changes in the `get_server_info`

bigmontz (approved):
🪂 

(Just a comment about tests and the PR description 🗡️)

PR #745, opened 2022-06-13
New timeout config options [ADR 006]

robsdedude:
Amends https://github.com/neo4j/neo4j-python-driver/pull/675

Implements [ADR 006](https://github.com/neo-technology/drivers-adr/blob/main/adrs/006-acquisition-timeouts.md) (internal reference)

PR #746, opened 2022-06-22
[4.4] Backport new timeout config options [ADR 006]

robsdedude:
Backport of https://github.com/neo4j/neo4j-python-driver/pull/745


neo4j/conf.py
```diff
@@ -256,6 +263,12 @@ class WorkspaceConfig(Config):
     """ WorkSpace configuration.
     """
 
+    #: Session Connection Timeout
+    session_connection_timeout = 120.0  # seconds
```
bigmontz: Shouldn't it be inf?

bigmontz (changes requested):
the default `session_connection_timeout` seems wrong.

neo4j/conf.py
robsdedude: It 100%... nah inf% should be! :open_mouth:

bigmontz (approved):
⏳

PR #749, opened 2022-06-29
UTC encoding patch for Bolt 4.4 and 4.3

robsdedude:
This is to be considered quite literally a patch.
A proper implementation of this functionality would require a lot of refactoring
that potentially involves moving modules and packages around that are
undocumented, hence they shouldn't be used, but they are not marked as internal
by a leading underscore. Therefore, I don't want to introduce such a refactoring
in a patch release of the driver.


bigmontz (approved):
🕥 🕦

PR #750, opened 2022-07-04
API docs: make simple example self-contained

robsdedude:
When demonstrating how to create nodes/relationships, the queries should entail
everything needed so that the example as is, does something useful on en empty
database.


AndyHeap-NeoTech (approved):
Looks good.

PR #759, opened 2022-07-12
Raise hydration errors on value access

robsdedude:
This has its limitations, thought. If a user returns for instance a list with one invalid value in it, the Record would throw on trying to access the list. So the user still could't salvage the other list entries. I don't think, however, there's a Pythonic way to cater for this use-case.

### DIFF
--- a/neo4j/__init__.py
+++ b/neo4j/__init__.py
@@ -108,7 +108,6 @@ __all__ = [
     "BoltDriver",
     "Bookmark",
     "Bookmarks",
-    "Config",
     "custom_auth",
     "DEFAULT_DATABASE",
     "Driver",
@@ -120,7 +119,6 @@ __all__ = [
     "kerberos_auth",
     "ManagedTransaction",
     "Neo4jDriver",
-    "PoolConfig",
     "Query",
     "READ_ACCESS",
     "Record",
@@ -128,7 +126,6 @@ __all__ = [
     "ResultSummary",
     "ServerInfo",
     "Session",
-    "SessionConfig",
     "SummaryCounters",
     "Transaction",
     "TRUST_ALL_CERTIFICATES",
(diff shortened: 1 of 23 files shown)


neo4j/__init__.py
```diff
@@ -108,7 +108,6 @@
     "BoltDriver",
     "Bookmark",
     "Bookmarks",
-    "Config",
```
bigmontz, line 111: Why does it get removed?

neo4j/_async/work/result.py
```diff
@@ -145,6 +146,11 @@ async def on_failed_attach(metadata):
     def _pull(self):
         def on_records(records):
             if not self._discarding:
+                records = (
+                    record.raw_data
+                    if isinstance(record, BrokenHydrationObject) else record
+                    for record in records
+                )
```
bigmontz, line 153: Why do we need this special behaviour?

bigmontz (approved):
🤖

neo4j/__init__.py
robsdedude: Guess I should put that into a separate PR, really. There is no reason why this class was exported in the first place. The user is not supposed to instantiate it, or ever receive an instance of it or anything. There is practically nothing the user can or should do with it.

neo4j/_async/work/result.py
robsdedude: That is because the `RECORD` message has one field which is the list of the record's values. That list will if any of its elements is a `BrokenHydrationObject` be reduced to a `BrokenHydrationObject` itself
https://github.com/neo4j/neo4j-python-driver/pull/759/files#diff-2a37374b359d1060498f6ecb133093a8ae46b21d00231fe42d8160ffc363e8edR68
Therefore, I need to unpack it for the `Record` class to be able to access the values.

neo4j/__init__.py
robsdedude: Well, actually, on the 5.0 dev branch, those exports are broken. While they are listed in `__all__`, they are not imported in `neo4j/__init__.py`. So this PR doesn't remove the export. It just fixes the already happened, but not finished up removal. So I think this is fine.

PR #761, opened 2022-07-19
Handle `asyncio.CancelledError` gracefully

robsdedude:


docs/source/async_api.rst
```diff
@@ -531,3 +570,48 @@ A :class:`neo4j.AsyncResult` is attached to an active connection, through a :cla
...
+        await session.close()
+
+As mentioned above, any cancellation of I/O work will cause the driver to close
+the affected connection. This will kill any :class:`neo4j.AsyncTransaction` and
+:class:`neo4j.AsyncResult` objects that are attached to that connection. Hence,
+after catching a :class:`asyncio.CancelledError`, you should not try to use
+transactions or results created earlier. They are likely to not be valid
+anymore.
+
+Furthermore, there is no a guarantee as to whether a piece of ongoing work got
```
robsdedude:
```suggestion
Furthermore, there is no guarantee as to whether a piece of ongoing work got
```

neo4j/_async/io/_pool.py
```diff
@@ -262,16 +301,33 @@ def in_use_connection_count(self, address):
...
+    async def _close_connections(cls, connections):
+        cancelled = None
+        for connection in connections:
+            if cancelled is not None:
+                connection.kill()
+                continue
+            try:
+                await connection.close()
+            except asyncio.CancelledError as e:
+                # We've got cancelled: not time to gracefully close these
```
robsdedude:
```suggestion
                # We've got cancelled: no more time to gracefully close these
```

neo4j/_async_compat/concurrency.py
```diff
@@ -108,7 +121,16 @@ async def acquire(self, blocking=True, timeout=-1):
...
         try:
-            await asyncio.wait_for(self._acquire(me), timeout)
+            fut = asyncio.ensure_future(self._acquire(me))
+            try:
+                await wait_for(fut, timeout)
+            except asyncio.CancelledError:
+                already_finished = not fut.cancel()
+                if already_finished:
+                    # Too late to cancel the acquisition.
+                    # This can only happen in Python 3.7's asyncio
```
robsdedude, line 131:
```suggestion
                    # This can only happen in Python 3.7's asyncio 
                    # as well as in our custom wait_for shim.
```

neo4j/_async/io/_bolt.py
```diff
@@ -746,18 +755,19 @@ async def close(self):
         finally:
             self._closed = True
 
-    async def close_non_blocking(self):
```
bigmontz, line 749: this is a public method, is there any other client for this method?

tests/integration/conftest.py
```diff
@@ -33,7 +33,7 @@ def run_and_rollback(tx, cypher, **parameters):
         raise ForcedRollback(value)
 
     def f(cypher, **parameters):
-        with bolt_driver.session() as session:
+        with driver.session() as session:
```
bigmontz, line 36: Why did it changed?

bigmontz (approved):
🌵 

(Just small comments)

neo4j/_async/io/_bolt.py
robsdedude: While `close_non_blocking` is public, the connection class is never exposed to the user. So there is no risk of breaking user-code. All driver-internal usages of `close_non_blocking` have been replace with `kill` which is the proper way of handling cancellations. I was also used to clean up failed or timed out connection attempts (error/timeout while sending/waiting for response to the `HELLO` message). I think it's also fine to forcefully close the connection in such cases.

tests/integration/conftest.py
robsdedude: The fixture was broken. `bolt_driver` is an undefined variable here. It just went unnoticed because this fixture is currently unused. But I don't want to get rid of it as it's a very handy one that I use every now and then (like I did while developing this PR, that's when I noticed). I also can't rule out that I might use this fixture for some tests at a later point.

neo4j/_async_compat/shims/__init__.py
```diff
@@ -0,0 +1,124 @@
...
+            except asyncio.CancelledError:
+                if fut.done():
+                    # [PATCH]
+                    # Applied patch to not swallow the outer cancellation.
+                    # See: https://github.com/python/cpython/pull/26097
+                    # and https://github.com/python/cpython/pull/28149
+
+                    # We got cancelled, but we are already done. Therefore,
+                    # we defer the cancellation until next time the task yields
+                    # to the event loop.
```
robsdedude: Comment does not match code any longer.

PR #763, opened 2022-07-20
Introduce API Redesign

bigmontz:
* [x] Query API
* [x] Execute API
* [x] Automatic Access Mode
* [x] Async
* [x] Docs


neo4j/_sync/work/result.py
```diff
@@ -707,3 +707,13 @@ def closed(self):
         .. versionadded:: 5.0
         """
         return self._out_of_scope or self._consumed
+
+
+class QueryResult:
```
robsdedude: Could be a `collections.namedtuple` https://docs.python.org/3/library/collections.html#collections.namedtuple

```python
from collections import namedtuple


QueryResult = namedtuple("QueryResult", ("records", "summary"))
```

Which has the advantage, that users can run
```python
records, _ = driver.run("<CYPHER>")
```

bigmontz: neat.

bigmontz:
Usage example:

```python
from importlib.metadata import metadata
import neo4j
import asyncio

from neo4j.api import CLUSTER_READERS_ACCESS, CLUSTER_WRITERS_ACCESS


async def main():
    driver = neo4j.AsyncGraphDatabase.driver(
        'bolt://localhost', auth=neo4j.basic_auth("neo4j", "pass"))

    async with driver.session() as session:
        async def txfunc(tx):
            query_result = await tx.query('RETURN $n as n', n=1)
            print(query_result.records)
            print(query_result.summary.database)

        await session.execute(txfunc)

    async with driver.session() as session:
        async def txfunc(tx):
            records, summary = await tx.query(
                'RETURN $n as n',
                n=2)
            print(records)
            print(summary.database)

        await session.execute(txfunc)

    async with driver.session() as session:
        records, summary = await session.query(
                'RETURN $n as n',
                n=3,
                skip_records=True,
                cluster_member_access=CLUSTER_WRITERS_ACCESS)
        print(records)
        print(summary.database)

    async with driver.session() as session:
        records, summary = await session.query(
                'RETURN $n as n',
                n=4)
        print(records)
        print(summary.database)

    async def txfunc(tx, n):
        records, summary = await tx.query(
                'RETURN $n as n',
                n=n)
        print(records)
        print(summary.database)

    # configuration are optional
    await driver.execute(
        lambda tx: txfunc(tx, n=5),
        database="neo4j",
        cluster_member_access=CLUSTER_WRITERS_ACCESS,
        timeout=10,
        metadata={"service": "python driver playground"}
    )

    records, summary = await driver.query(
      'RETURN $n as n',
      n=6,
      cluster_member_access=CLUSTER_READERS_ACCESS)

    print(records)
    print(summary.database)

    await driver.close()


asyncio.run(main())
```

neo4j/_async/driver.py
```diff
@@ -392,6 +392,89 @@ async def supports_multi_db(self):
...
+
+    async def execute(self, transaction_function, *args, **kwargs):
+        """Execute a unit of work in a managed transaction.
+
+        .. note::
+            This does not necessarily imply access control, see the session
+            configuration option :ref:`default-access-mode-ref`.
+
+        This transaction will automatically be committed unless an exception is thrown during query execution or by the user code.
+        Note, that this function perform retries and that the supplied `transaction_function` might get invoked more than once.
```
robsdedude: Code-style: max 79 chars/line

neo4j/_async/driver.py
```diff
@@ -392,6 +392,89 @@ async def supports_multi_db(self):
...
+
+        Example::
+
+            async def do_cypher_tx(tx):
+                records, _ = await tx.query("RETURN 1 AS x")
+                return records
+
+            values = await driver.execute(do_cypher_tx,
+                database="neo4j",
+                cluster_member_access=neo4j.api.CLUSTER_READERS_ACCESS)
```
robsdedude:
```suggestion
            values = await driver.execute(
                do_cypher_tx,
                database="neo4j",
                cluster_member_access=neo4j.api.CLUSTER_READERS_ACCESS
            )
```

neo4j/_async/work/session.py
```diff
@@ -40,7 +43,7 @@
     TransactionError,
 )
 from ...work import Query
-from .result import AsyncResult
+from .result import AsyncResult, QueryResult
```
robsdedude: Please install the pre-commit hook as explained [here](https://github.com/neo4j/neo4j-python-driver/blob/5.0/CONTRIBUTING.md#specifically-for-this-project).

It would've auto-formatted this import properly for you :)

```python
from .result import (
    AsyncResult,
    QueryResult,
)
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +242,106 @@ async def run(self, query, parameters=None, **kwargs):
...
+
+        Example::
+
+            async def do_cypher_tx(tx):
+                records, _ = await tx.query("RETURN 1 AS x")
+                return records
+
+            async with driver.session() as session:
+                values = await session.execute(do_cypher_tx,
+                    cluster_member_access=neo4j.api.CLUSTER_READERS_ACCESS)
```
robsdedude:
```suggestion
                values = await session.execute(
                    do_cypher_tx,
                    cluster_member_access=neo4j.api.CLUSTER_READERS_ACCESS
                )
```

neo4j/_async/work/transaction.py
```diff
@@ -22,7 +22,7 @@
 from ...exceptions import TransactionError
 from ...work import Query
 from ..io import ConnectionErrorHandler
-from .result import AsyncResult
+from .result import AsyncResult, QueryResult
```
robsdedude: import style (pre-commit hook will fix)

neo4j/_async/work/session.py
```diff
@@ -239,6 +242,106 @@ async def run(self, query, parameters=None, **kwargs):
...
 
+    async def query(self, query, parameters=None, **kwargs):
+        """
+        :param query: cypher query
+        :type query: str, neo4j.Query
+        :param parameters: dictionary of parameters
+        :type parameters: dict
+        :param kwargs: additional keyword parameters
+        :returns: a new :class:`neo4j.QueryResult` object
+        :rtype: QueryResult
```
robsdedude, line 292: I like to separate param docs, return docs, and raises docs each by a blank line for better readability.

```suggestion

        :returns: a new :class:`neo4j.QueryResult` object
        :rtype: QueryResult
```

robsdedude: Applies to other places as well, won't comment.

neo4j/_async/work/session.py
```diff
@@ -239,6 +242,109 @@ async def run(self, query, parameters=None, **kwargs):
...
+        :return: a result as returned by the given unit of work
+        """
+        cluster_member_access = kwargs.pop(
+            "cluster_member_access", CLUSTER_AUTO_ACCESS)
+
+        if cluster_member_access == CLUSTER_AUTO_ACCESS:
+            if await self._supports_auto_routing():
+                access_mode = READ_ACCESS
+            else:
+                raise ValueError('Server does not support CLUSTER_AUTO_ACCESS')
```
robsdedude: There is no wrong or right but I like to keep it consistent and I decided to go with the double quotes.
```suggestion
                raise ValueError("Server does not support CLUSTER_AUTO_ACCESS")
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +242,109 @@ async def run(self, query, parameters=None, **kwargs):
...
+        :param transaction_function: a function that takes a transaction as an
+            argument and does work with the transaction.
+            `transaction_function(tx, *args, **kwargs)` where `tx` is a
+            :class:`.Transaction`.
+        :param args: arguments for the `transaction_function`
+        :param kwargs: key word arguments for the `transaction_function`
+        :return: a result as returned by the given unit of work
+        """
+        cluster_member_access = kwargs.pop(
+            "cluster_member_access", CLUSTER_AUTO_ACCESS)
```
robsdedude:
```suggestion
        cluster_member_access = kwargs.pop(
            "cluster_member_access", CLUSTER_AUTO_ACCESS
        )
```
This is not particularly PEP8 style-guide but personal preference on how I like to do my parenthesis. If I open them as last character on a line, I have the matching closing parentheses on its own line. More verbose but I find it easier to see where the call ends.

neo4j/_async/work/workspace.py
```diff
@@ -125,6 +127,16 @@ async def _disconnect(self, sync=False):
                 self._connection = None
             self._connection_access_mode = None
 
+    async def _supports_auto_routing(self):
+        if self._pool.is_direct():
+            return True
+
+        await self._connect(READ_ACCESS)
+        supports_auto_routing = self._connection.configuration_hints.get(
+            'server_side_routing', False)
```
robsdedude:
```suggestion
        supports_auto_routing = self._connection.configuration_hints.get(
            "server_side_routing", False
        )
```

neo4j/_async/work/session.py
bigmontz: It is installed, but the changes was mad previous the installation.

neo4j/_sync/driver.py
```diff
@@ -392,6 +392,89 @@ def supports_multi_db(self):
...
+
+
+            values = driver.execute(get_two_tx)
+
+        :param transaction_function: a function that takes a transaction as an
+            argument and does work with the transaction.
+            `transaction_function(tx, *args, **kwargs)` where `tx` is a
+            :class:`.Transaction`.
+        :param args: arguments for the `transaction_function`
+        :param kwargs: key word arguments for the `transaction_function`
```
robsdedude: The docs use [reStructuredText](https://www.writethedocs.org/guide/writing/reStructuredText/) not markdown (I wish they would :upside_down_face:)
```suggestion
            argument and does work with the transaction.
            ``transaction_function(tx, *args, **kwargs)`` where ``tx`` is a
            :class:`.Transaction`.
        :param args: arguments for the ``transaction_function``
        :param kwargs: key word arguments for the ``transaction_function``
```

neo4j/_async/driver.py
```diff
@@ -392,6 +392,91 @@ async def supports_multi_db(self):
...
+
+            
+            values = await driver.execute(get_two_tx)
+
+        :param transaction_function: a function that takes a transaction as an
+            argument and does work with the transaction.
+            `transaction_function(tx, *args, **kwargs)` where `tx` is a
+            :class:`.Transaction`.
+        :param args: arguments for the `transaction_function`
+        :param kwargs: key word arguments for the `transaction_function`
```
robsdedude:
```suggestion
            ``transaction_function(tx, *args, **kwargs)`` where ``tx`` is a
            :class:`.Transaction`.
        :param args: arguments for the ``transaction_function``
        :param kwargs: key word arguments for the ``transaction_function``
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +242,109 @@ async def run(self, query, parameters=None, **kwargs):
...
+
+    async def execute(self, transaction_function, *args, **kwargs):
+        """Execute a unit of work in a managed transaction.
+
+        .. note::
+            This does not necessarily imply access control, see the session
+            configuration option :ref:`default-access-mode-ref`.
+
+        This transaction will automatically be committed unless an exception is thrown during query execution or by the user code.
+        Note, that this function perform retries and that the supplied `transaction_function` might get invoked more than once.
```
robsdedude:
```suggestion
        Note, that this function perform retries and that the supplied ``transaction_function`` might get invoked more than once.
```

neo4j/_async/driver.py
```diff
@@ -392,6 +392,91 @@ async def supports_multi_db(self):
...
+
+    async def execute(self, transaction_function, *args, **kwargs):
+        """Execute a unit of work in a managed transaction.
+
+        .. note::
+            This does not necessarily imply access control, see the session
+            configuration option :ref:`default-access-mode-ref`.
+
+        This transaction will automatically be committed unless an exception is thrown during query execution or by the user code.
+        Note, that this function perform retries and that the supplied `transaction_function` might get invoked more than once.
```
robsdedude:
```suggestion
        Note, that this function perform retries and that the supplied ``transaction_function`` might get invoked more than once.
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +242,109 @@ async def run(self, query, parameters=None, **kwargs):
...
+
+            async with driver.session() as session:
+                values = await session.execute(get_two_tx)
+
+        :param transaction_function: a function that takes a transaction as an
+            argument and does work with the transaction.
+            `transaction_function(tx, *args, **kwargs)` where `tx` is a
+            :class:`.Transaction`.
+        :param args: arguments for the `transaction_function`
+        :param kwargs: key word arguments for the `transaction_function`
```
robsdedude:
```suggestion
            ``transaction_function(tx, *args, **kwargs)`` where ``tx`` is a
            :class:`.Transaction`.
        :param args: arguments for the ``transaction_function``
        :param kwargs: key word arguments for the ``transaction_function``
```

neo4j/_sync/io/_pool.py
```diff
@@ -348,6 +348,9 @@ def acquire(
             self.address, deadline, liveness_check_timeout
         )
 
+    def is_direct(self):
```
robsdedude, line 351: I like this method.

 * I think the base Pool class should have it too and should either `raise NotImplementedError` or become a `abc.ABC` (abstract base class).
 * There are already places in the driver that check whether the pool is direct or not (currently using `isinstance(pool, Neo4jPool)` or similar) . They could also be adopted to use this method.
 * It could also be an attribute just because the lookup might be ever so slightly faster than a method call. But that might be a bit much. Yeah whatever, that one was just a quick thought.

neo4j/_async/work/session.py
bigmontz: Funny thing, I also prefer the way you suggested. However, I used to work in a repository which the default was the way I wrote

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
...
+        The query is sent and the result header received
+        immediately and the :class:`neo4j.QueryResult`is
+        fetched.
+
+        For more usage details, see :meth:`.AsyncTransaction.query`.
+
+        For auto-commit queries, use `AsyncSession.run`.
+
+        For access to the neo4j.AsyncResult object,
+        use `AsyncDriver.execute` and `.AsyncTransaction.run`
```
bigmontz:
```suggestion
        For auto-commit queries, use :meth:`AsyncSession.run`.

        For access to the :class:`neo4j.AsyncResult` object,
        use :meth:`AsyncDriver.execute` and :meth:`.AsyncTransaction.run`
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+        The query is sent and the result header received
+        immediately and the :class:`neo4j.QueryResult`is
+        fetched.
+
+        For more usage details, see :meth:`.AsyncTransaction.query`.
+
+        For auto-commit queries, use `AsyncSession.run`.
+
+        For access to the neo4j.AsyncResult object,
+        use `AsyncSession.execute` and `.AsyncTransaction.run`
```
bigmontz:
```suggestion
        For more usage details, see :meth:`.AsyncTransaction.query`.

        For auto-commit queries, use :class:`AsyncSession.run`.

        For access to the :class:`neo4j.AsyncResult` object,
        use :meth:`AsyncSession.execute` and :meth:`.AsyncTransaction.run`
```

neo4j/_sync/work/session.py
```diff
@@ -239,6 +245,125 @@ def run(self, query, parameters=None, **kwargs):
...
+        The query is sent and the result header received
+        immediately and the :class:`neo4j.QueryResult`is
+        fetched.
+
+        For more usage details, see :meth:`.Transaction.query`.
+
+        For auto-commit queries, use `Session.run`.
+
+        For access to the neo4j.Result object,
+        use `Session.execute` and `.Transaction.run`
```
bigmontz:
```suggestion
        The query is sent and the result header received
        immediately and the :class:`neo4j.QueryResult`is
        fetched.

        For more usage details, see :meth:`.Transaction.query`.

        For auto-commit queries, use :meth:`Session.run`.

        For access to the :class:`neo4j.Result` object,
        use :meth:`Session.execute` and :meth:`.Transaction.run`
```

neo4j/_sync/driver.py
```diff
@@ -392,6 +392,108 @@ def supports_multi_db(self):
...
+        The query is sent and the result header received
+        immediately and the :class:`neo4j.QueryResult`is
+        fetched.
+
+        For more usage details, see :meth:`.Transaction.query`.
+
+        For auto-commit queries, use `Session.run`.
+
+        For access to the neo4j.Result object,
+        use `Driver.execute` and `.Transaction.run`
```
bigmontz:
```suggestion
        The query is sent and the result header received
        immediately and the :class:`neo4j.QueryResult`is
        fetched.

        For more usage details, see :meth:`.Transaction.query`.

        For auto-commit queries, use :meth:`Session.run`.

        For access to the :class:`neo4j.Result` object,
        use :meth:`Driver.execute` and :meth:`.Transaction.run`
```

neo4j/_async/work/transaction.py
```diff
@@ -131,6 +134,43 @@ async def run(self, query, parameters=None, **kwparameters):
...
+
+        :returns: a new :class:`neo4j.QueryResult` object
+        :rtype: :class:`neo4j.QueryResult`
+
+        :raise TransactionError: if the transaction is already closed
+        """
+        result = await self.run(query, parameters, **kwparameters)
+        records = []
+        async for x in result:
+            records.append(x)
```
robsdedude:
```suggestion
        records = await AsyncUtil.list(result)
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+            return await tx.query(query, parameters, **job_kwargs)
+
+        return await self.execute(job, **kwargs)
+
+    async def execute(self, transaction_function, *args, **kwargs):
+        """Execute a unit of work in a managed transaction.
+
+        .. note::
+            This does not necessarily imply access control, see the session
+            configuration option :ref:`default-access-mode-ref`.
```
robsdedude: This note comes out of the blue. The reader will (at least it did) wonder what "This" refers to.

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+        """
+        cluster_member_access = kwargs.pop(
+            "cluster_member_access", CLUSTER_AUTO_ACCESS
+        )
+
+        if cluster_member_access == CLUSTER_AUTO_ACCESS:
+            if await self._supports_auto_routing():
+                access_mode = READ_ACCESS
+            else:
+                raise ValueError("Server does not support CLUSTER_AUTO_ACCESS")
```
robsdedude: There is a `ConfigurationError` in the driver that is usually thrown when the remote doesn't support a certain action/option.

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+        elif cluster_member_access == CLUSTER_READERS_ACCESS:
+            access_mode = READ_ACCESS
+        elif cluster_member_access == CLUSTER_WRITERS_ACCESS:
+            access_mode = WRITE_ACCESS
+        else:
+            raise ValueError("Invalid cluster_member_access")
```
robsdedude: While `ValueError` is a reasonable choice (I would've probably chosen it if there was no driver already), I checked the existing code to see what gets raised when configuring a session with an invalid access mode: it's `ClientError`. So I'd say we should raise that for more consistency.

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
 
         return self._auto_result
 
+    async def query(self, query, parameters=None, **kwargs):
+        """
+        Run a Cypher query within an managed transaction and
+        all the retries policy will be applied.
```
robsdedude: The first paragraph of a doc string should be one line only.

If you want to did deeper https://peps.python.org/pep-0008/#documentation-strings and https://peps.python.org/pep-0257/

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+
+        The query is sent and the result header received
+        immediately and the :class:`neo4j.QueryResult`is
+        fetched.
```
robsdedude: I don't really know what the result header is supposed to mean. If I had to guess, I's say it's the first `SUCCESS` message that contains the `fields` fields. If so, that's way too technical and in detail for the end-user.
```suggestion
        The query is sent and the full result is fetched and returned as
        :class:`neo4j.QueryResult`.
```

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
...
+
+            values = await driver.execute(get_two_tx)
+
+        :param transaction_function: a function that takes a transaction as an
+            argument and does work with the transaction.
+            ``transaction_function(tx, *args, **kwargs)`` where ``tx`` is a
+            :class:`.Transaction`.
+        :param args: arguments for the ``transaction_function``
+        :param kwargs: key word arguments for the ``transaction_function``
+        :return: a result as returned by the given unit of work
```
robsdedude, line 534:
```suggestion

        :return: a result as returned by the given unit of work
```

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+            async with driver.session() as session:
+                values = await session.execute(get_two_tx)
+
+        :param transaction_function: a function that takes a transaction as an
+            argument and does work with the transaction.
+            ``transaction_function(tx, *args, **kwargs)`` where ``tx`` is a
+            :class:`.Transaction`.
+        :param args: arguments for the ``transaction_function``
+        :param kwargs: key word arguments for the ``transaction_function``
+        :return: a result as returned by the given unit of work
```
robsdedude, line 386:
```suggestion

        :return: a result as returned by the given unit of work
```

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
             await session._connect(READ_ACCESS)
             return session._connection.supports_multiple_databases
 
+    async def query(self, query, parameters=None, **kwargs):
+        """
+        Run a Cypher query within an managed transaction and
+        all the retries policy will be applied.
```
robsdedude: One line only, please :)

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
...
+        :type query: str, neo4j.Query
+        :param parameters: dictionary of parameters
+        :type parameters: dict
+        :param kwargs: additional keyword parameters
+
+        :returns: a new :class:`neo4j.QueryResult` object
+        :rtype: QueryResult
+        """
+        session_kwargs = {}
+        if "database" in kwargs:
```
robsdedude: The doc string doesn't mention anywhere that the `database` kwarg has a special function and will not get used as a query parameter. Maybe, `database=None` should be an explicit keyword argument, as in

```python
async def query(self, query, parameters=None, database=None, **kwargs):
```

Tbf, this also applies to `driver.session`. It does not specify what config options there are, but at least there is an explicit list of available config options in the API docs https://github.com/neo4j/neo4j-python-driver/blob/5.0/docs/source/api.rst#session-configuration

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
...
+
+        async with self.session(**session_kwargs) as session:
+            return await session.query(query, parameters, **kwargs)
+
+    async def execute(self, transaction_function, *args, **kwargs):
+        """Execute a unit of work in a managed transaction.
+
+        .. note::
+            This does not necessarily imply access control, see the session
+            configuration option :ref:`default-access-mode-ref`.
```
robsdedude: Out of the blue (see other comment)

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
...
+
+        This transaction will automatically be committed unless an exception
+        is thrown during query execution or by the user code.
+        Note, that this function perform retries and that the supplied `
+        transaction_function` might get invoked more than once.
```
robsdedude:
```suggestion
        Note, that this function perform retries and that the supplied
        ``transaction_function`` might get invoked more than once.
```

neo4j/_async/driver.py
```diff
@@ -392,6 +392,108 @@ async def supports_multi_db(self):
...
+
+        Managed transactions should not generally be explicitly committed
+        (via ``tx.commit()``).
```
robsdedude:
```suggestion
```
It's funny that this is still in the docs :hand_over_mouth: in fact, they can't anymore. This method does no longer exist for managed transactions.

neo4j/_async/work/session.py
```diff
@@ -239,6 +245,126 @@ async def run(self, query, parameters=None, **kwargs):
...
+            This does not necessarily imply access control, see the session
+            configuration option :ref:`default-access-mode-ref`.
+
+        This transaction will automatically be committed unless an exception
+        is thrown during query execution or by the user code.
+        Note, that this function perform retries and that the supplied
+        ``transaction_function`` might get invoked more than once.
+
+        Managed transactions should not generally be explicitly committed
+        (via ``tx.commit()``).
```
robsdedude:
```suggestion
```

neo4j/_async/work/transaction.py
```diff
@@ -131,6 +134,43 @@ async def run(self, query, parameters=None, **kwparameters):
...
+        :class:`str`, :class:`list` and :class:`dict`. Note however that
+        :class:`list` properties must be homogenous.
+
+        :param query: cypher query
+        :type query: str
+        :param parameters: dictionary of parameters
+        :type parameters: dict
+        :param kwparameters: additional keyword parameters
+
+        :returns: a new :class:`neo4j.QueryResult` object
```
robsdedude: The type is already documented below :upside_down_face: 
```suggestion
        :returns: the result of the query
```

neo4j/_async/work/session.py
bigmontz: Out of the ctrl c + ctrl v

neo4j/_async/driver.py
```diff
@@ -416,14 +421,20 @@ async def query(self, query, parameters=None, **kwargs):
...
-        session_kwargs = {}
-        if "database" in kwargs:
-            session_kwargs["database"] = kwargs.pop("database")
 
-        async with self.session(**session_kwargs) as session:
-            return await session.query(query, parameters, **kwargs)
+        async with self.session(database=database) as session:
+            return await session.query(
+                query,
+                parameters,
```
robsdedude:
```suggestion
                parameters=parameters,
```

That's generally more future-proof.

neo4j/_async/work/session.py
```diff
@@ -278,9 +280,15 @@ async def job(tx, **job_kwargs):
...
+        return await self.execute(
+            job,
+            cluster_member_access=cluster_member_access,
+            **kwargs
+        )
 
-    async def execute(self, transaction_function, *args, **kwargs):
+    async def execute(self, transaction_function, *args,
+                      cluster_member_access=CLUSTER_AUTO_ACCESS,
+                      **kwargs):
```
robsdedude: 1. about the code-style. This is fine, but when run through unasync will look like this
```python
    def execute(self, transaction_function, *args,
                      cluster_member_access=CLUSTER_AUTO_ACCESS,
                      **kwargs):
```
:nauseated_face: 

This can be fixed like so
```suggestion
    async def execute(
        self, transaction_function, *args, 
        cluster_member_access=CLUSTER_AUTO_ACCESS, **kwargs
    ):
```

There are other places like this. Basically whenever you have an async function, with multiple lines of arguments, put the first argument on a new line. Note that this is not general Python style-guide, but just necessary to make the output of unasync look neat.

robsdedude: 2. I don't like this design. Here is the problem: if you write a wrapper function that does not control the signature of `transaction_function`, you are screwed, because you can't make it right in the general case because `transaction_function` could have a keyword argument `cluster_member_access`. What now?

This could be addressed like so
```python
    async def execute(
        self, transaction_function,
        transaction_function_args=None, transaction_function_kwargs=None,
        cluster_member_access=CLUSTER_AUTO_ACCESS
    ):
```
which is admittedly less nice to use. There is the third option
```python
    async def execute(
        self, transaction_function, *args,
        transaction_function_args=None, transaction_function_kwargs=None,
        cluster_member_access=CLUSTER_AUTO_ACCESS, **kwargs
    ):
```
But I'm not sure if that's not just more confusing than it's helping. I really don't know. I think this is the major downside of cramming all our API into easy to access functions... They become bloated, more complicated, and have to do many things at once.

bigmontz: The `cluster_member_access` was already being not passed to the tx function in the previous version and it is kind of useless in that context, since none of the tx methods accept this argument.


I've tried to keep usage quite simple while keep the usage quite similar to the previous tx function. The difference is now the param which will be removed is documented.

An alternative is removing the extra arguments (*args and *kwargs), this way the tx function will only receive the transaction.

robsdedude: I don't understand. Isn't `transaction_function` a function defined by user-level code (outside of the driver)? How can you know it doesn't have a `cluster_member_access` parameter?

neo4j/_async/driver.py
```diff
@@ -393,11 +393,11 @@ async def supports_multi_db(self):
...
-    async def query(self, query, parameters=None,
-                    database=None,
-                    cluster_member_access=CLUSTER_AUTO_ACCESS,
-                    skip_records=False,
-                    **kwargs):
+    async def query(
+            self, query, parameters=None, database=None,
+            cluster_member_access=CLUSTER_AUTO_ACCESS, skip_records=False,
+            **kwargs
+    ):
```
robsdedude:
```suggestion
    async def query(
        self, query, parameters=None, database=None,
        cluster_member_access=CLUSTER_AUTO_ACCESS, skip_records=False,
        **kwargs
    ):
```

neo4j/_async/driver.py
```diff
@@ -431,10 +431,11 @@ async def query(self, query, parameters=None,
...
             )
 
-    async def execute(self, transaction_function, *args,
-                      database=None,
-                      cluster_member_access=CLUSTER_AUTO_ACCESS,
-                      **kwargs):
+    async def execute(
+            self, transaction_function, *args,
+            database=None, cluster_member_access=CLUSTER_AUTO_ACCESS,
+            **kwargs
```
robsdedude:
```suggestion
        self, transaction_function, *args,
        database=None, cluster_member_access=CLUSTER_AUTO_ACCESS,
        **kwargs
```

neo4j/_sync/driver.py
```diff
@@ -393,11 +393,11 @@ def supports_multi_db(self):
...
 
-    def query(self, query, parameters=None,
-                    database=None,
-                    cluster_member_access=CLUSTER_AUTO_ACCESS,
-                    skip_records=False,
-                    **kwargs):
+    def query(
+            self, query, parameters=None, database=None,
+            cluster_member_access=CLUSTER_AUTO_ACCESS, skip_records=False,
+            **kwargs
```
robsdedude:
```suggestion
        self, query, parameters=None, database=None,
        cluster_member_access=CLUSTER_AUTO_ACCESS, skip_records=False,
        **kwargs
```

neo4j/_async/work/session.py
bigmontz: We don't know. The `Session.run` also have the same problem, the user couldn't have a parameter called `parameters` (if they set the param like `session.run('return n, parameters', n=1, parameters='a')`).

We could also always add this name params to the `tx_func` kwargs. Which is kind of the approach used by to extract `metadata` and `timeout` in `read_transaction`.

robsdedude: While this is true about the `session.run`, if I was to write a wrapper, I'd just never call `session.run(query, **parameters)` but instead `session.run(query, parameters=parameters)`. That way, I can be sure that my wrapper can cope with any query parameters thrown at it.

> We could also always add this name params to the tx_func kwargs. Which is kind of the approach used by to extract metadata and timeout in read_transaction.

I can't follow, sorry.

bigmontz: `read_transaction` and `write_transaction` get `timeout` and `metadata` from the `**kwargs` without remove it from it. We could do it with the `cluster_member_access` (and `database`, in the `driver.execute` case) by put it back to the `**kwargs`. However, we will have the issue with the documentation of the params. 

For solving this, I'm suggesting keep the signature like:

```python
async def execute(
        self, transaction_function, *args,
        cluster_member_access=None, **kwargs
    ):
```

Then, in the execute implementation do something like:

```python
# user doesn't set anything
if cluster_member_access is None:
    cluster_member_access = CLUSTER_AUTO_ACCESS
# user set something
else:
    kwargs['cluster_member_access'] = cluster_member_access
```

The problem is: we don't know if the user set named param `cluster_member_access` or not, we only can guess if they are using the default value.

bigmontz: There is still another issue: the user could use a different meaning and dataset for the `cluster_member_access` variable.

robsdedude:
[ADR](https://github.com/neo4j/neo4j-python-driver/pull/763) (internal link) was abandoned.

`driver.execute_query` (https://github.com/neo4j/neo4j-python-driver/pull/833) is it's spiritual successor.

PR #767, opened 2022-07-28
Type Hints

robsdedude:


neo4j/_async/driver.py
```diff
@@ -33,125 +43,176 @@
...
+        @AsyncUtil.experimental_async(
+            "neo4j async is in experimental phase. It might be removed or "
+            "changed at any time (including patch releases)."
         )
+        def driver(
+            cls,
+            uri: str,
+            *,
+            auth: t.Union[t.Tuple[t.Any, t.Any], Auth] = None,
+            **config  # TODO: type config
```
bigmontz: It seems to by typed in the other branch of this `if/else`

neo4j/exceptions.py
```diff
@@ -68,15 +68,47 @@
...
+
+    _T_Transaction = t.Union[AsyncManagedTransaction, AsyncTransaction,
+                             ManagedTransaction, Transaction]
+    _T_Result = t.Union[AsyncResult, Result]
+    _T_Session = t.Union[AsyncSession, Session]
+else:
+    _T_Transaction = t.Union["AsyncManagedTransaction", "AsyncTransaction",
+                             "ManagedTransaction", "Transaction"]
+    _T_Result = t.Union["AsyncResult", "Result"]
+    _T_Session = t.Union["AsyncSession", "Session"]
```
bigmontz, line 100: Why does it need to have different treatment for `TYPE_CHECKING` and regular code in this specific scenario?

neo4j/spatial/__init__.py
```diff
@@ -72,22 +72,6 @@ def dehydrate_point(value):
...
-def dehydrate_point(value):
-    """ Dehydrator for Point data.
-
-    :param value:
-    :type value: Point
-    :return:
-    """
-    return _hydration.dehydrate_point(value)
-
-
```
bigmontz, line 90: Why did it get removed before 6.0?

bigmontz (approved):
🔆

neo4j/exceptions.py
robsdedude: Because import `._async.work`/`._sync.work` causes a cyclic dependency here: :boom: 
So I use a forward reference (done with strings) if not type checking. This is actually only necessary to make the docs tool pick up the correct types when I use `_T_Transaction`, etc. as return/parameter type.

I could probably get away with only using forward referencing (in both cases) :shrug:

neo4j/spatial/__init__.py
robsdedude: Because I'm a derp :sweat_smile: 

That was a copy pasta mistake. The exact same code can be found right above it https://github.com/neo4j/neo4j-python-driver/pull/767/files/a2b0ef051806e05cb2b2cf5c63a31e318dd51c69#diff-0260b4d7c5791c8c36f46ad24ae30e20fff2a48958b9c918b59be77428be9864L59-L72

bigmontz: I marked the files which I reviewed as viewed, so the link you send is not working properly. Could you tell me the name of the file? (because, the lines I can guess by the line, L59-L72)

robsdedude: `neo4j/spatial/__init__.py` (the same file, right above the removed code you highlighted)

PR #771, opened 2022-08-03
[ADR-001] bookmark manager

robsdedude:
To do:
 - [x] test against https://github.com/neo4j-drivers/testkit/pull/502
 - [x] test against https://github.com/neo4j-drivers/testkit/pull/506
 - [x] write unit tests


bin/make-unasync
```diff
@@ -154,12 +154,12 @@ class CustomRule(unasync.Rule):
                     start += 1
                     end += 1
                 else:
-                    out += self._unasync_prefix(name[start:(end - 1)])
+                    out += self._unasync_name(name[start:(end - 1)])
```
bigmontz, line 157: Out of curiosity, what does it change?

docs/source/api.rst
```diff
@@ -138,7 +139,7 @@ Alternatively, one of the auth token helper functions can be used.
 Driver
 ******
 
-Every Neo4j-backed application will require a :class:`neo4j.Driver` object.
+Every Neo4j-backed application will require a driver object.
```
bigmontz, line 142: Why was the class reference removed?

bin/make-unasync
fbiville, line 157: was sind diese?

docs/source/api.rst
```diff
@@ -417,6 +419,29 @@ Specify the client agent name.
 :Default: *The Python Driver will generate a user agent name.*
 
 
+.. _bookmark-manager-ref:
+
+``bookmark_manager``
+--------------------
+Specify a bookmark manager for the driver to use. If present, the bookmark
+manger is used to keep all work on the driver causally consistent.
```
fbiville:
```suggestion
manager is used to keep all work on the driver causally consistent.
```
were you hungry when you wrote this? ;D

fbiville: nitpicking: it will be causally consistent iff one does not write/use bad BM implementations

docs/source/api.rst
```diff
@@ -825,7 +850,7 @@ Returning a live result object would prevent the driver from correctly managing
 
 This function will receive a :class:`neo4j.ManagedTransaction` object as its first parameter.
 
-.. autoclass:: neo4j.ManagedTransaction
+.. autoclass:: neo4j.ManagedTransaction()
```
fbiville, line 862: why the parentheses here? `GraphDatabase` does not have them

docs/source/async_api.rst
```diff
@@ -505,7 +534,7 @@ Returning a live result object would prevent the driver from correctly managing
 
 This function will receive a :class:`neo4j.AsyncManagedTransaction` object as its first parameter.
 
-.. autoclass:: neo4j.AsyncManagedTransaction
+.. autoclass:: neo4j.AsyncManagedTransaction()
```
fbiville, line 535: same parenthesis question

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+
+from .._async_compat.concurrency import AsyncCooperativeLock
+from .._async_compat.util import AsyncUtil
+from ..api import (
+    AsyncBookmarkManager,
+    Bookmarks,
+)
+
+
+T_BmSupplier = t.Callable[[t.Optional[str]],
```
fbiville, line 32: oh, you used Optional to avoid have an interface with 2 methods or having 2 callables, is that right?

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+                          t.Union[Bookmarks, t.Awaitable[Bookmarks]]]
```
fbiville, line 33: since you're in the async bookmark manager, shouldn't it always return `t.Awaitable[Bookmarks]`? why the union here?

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+T_BmConsumer = t.Callable[[str, Bookmarks], t.Union[None, t.Awaitable[None]]]
```
fbiville, line 34: same question as above

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+        self._lock = AsyncCooperativeLock()
+
+    async def update_bookmarks(
+        self, database: str, previous_bookmarks: t.Iterable[str],
+        new_bookmarks: t.Iterable[str]
+    ) -> None:
+        new_bms = set(new_bookmarks)
+        prev_bms = set(previous_bookmarks)
+        with self._lock:
+            if not new_bms:
```
fbiville: ah, I did not even check in Go that the new bookmarks could be empty :|

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+                return
+            curr_bms = self._bookmarks[database]
+            curr_bms.difference_update(prev_bms)
```
fbiville: could `difference_update` just be called `remove` or `remove_all`?

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+            )
+
+    async def get_bookmarks(self, database: str) -> t.Set[str]:
+        with self._lock:
+            bms = set(self._bookmarks[database])
+        if self._bookmarks_supplier:
+            extra_bms = await AsyncUtil.callback(
+                self._bookmarks_supplier, database
+            )
+            bms.update(extra_bms.raw_values)
```
fbiville, line 88: does this call alter the bookmarks stored in the `self._bookmarks`? In Go, it would, so I explicitly copy the bookmarks before the supplied ones get added to them

neo4j/_async/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+        self._lock = AsyncCooperativeLock()
+
+    async def update_bookmarks(
+        self, database: str, previous_bookmarks: t.Iterable[str],
+        new_bookmarks: t.Iterable[str]
+    ) -> None:
+        new_bms = set(new_bookmarks)
+        prev_bms = set(previous_bookmarks)
+        with self._lock:
```
fbiville, line 70: so you went with a simple lock, right? not a rwlock, it seems, or?

neo4j/_async/driver.py
```diff
@@ -62,11 +64,16 @@
...
+)
 from .work import AsyncSession
 
 
 class AsyncGraphDatabase:
-    """Accessor for :class:`neo4j.Driver` construction.
+    """Accessor for :class:`neo4j.AsyncDriver` construction.
     """
 
     if t.TYPE_CHECKING:
```
fbiville, line 79: oh, you have to explicitly check for this?

neo4j/_async/driver.py
```diff
@@ -206,6 +215,61 @@ def driver(cls, uri, *, auth=None, **config) -> AsyncDriver:
...
+
+            driver = neo4j.AsyncGraphDatabase.driver(
+                uri, auth=..., # ...
+                bookmark_manager=neo4j.AsyncGraphDatabase.bookmark_manager(
+                    # ... configure the bookmark manager
+                )
+            )
+
+        :param initial_bookmarks:
+            The initial set of bookmarks. The default bookmark manager will
```
fbiville:
```suggestion
            The initial set of bookmarks. The built-in implementation of the bookmark manager will
```
just saying "The default bookmark manager" may leave people to think it's enabled by default.
The default bookmark manager is currently a no-op.

neo4j/_async/driver.py
```diff
@@ -343,13 +407,16 @@ def session(
...
             bookmarks: t.Union[t.Iterable[str], Bookmarks, None] = ...,
+            ignore_bookmark_manager: bool = ...,
             default_access_mode: str = ...,
 
             # undocumented/unsupported options
             # they may be change or removed any time without prior notice
             initial_retry_delay: float = ...,
             retry_delay_multiplier: float = ...,
             retry_delay_jitter_factor: float = ...,
+            bookmark_manager: t.Union[AsyncBookmarkManager,
```
fbiville: why is the bookmark manager exposed here and not at the driver level?

fbiville: ah, maybe it's not the public API to create a session from the driver, or is it?

neo4j/_async_compat/concurrency.py
```diff
@@ -423,6 +436,8 @@ def notify_all(self):
         self.notify(len(self._waiters))
 
 
-Condition = threading.Condition
-CooperativeLock = Lock = threading.Lock
-CooperativeRLock = RLock = threading.RLock
+Condition: te.TypeAlias = threading.Condition
+CooperativeLock: te.TypeAlias = threading.Lock
+Lock: te.TypeAlias = threading.Lock
+CooperativeRLock: te.TypeAlias = threading.RLock
```
fbiville, line 442: you're not using this or the one below, or did I miss something?

tests/unit/async_/io/test_class_bolt4x3.py
```diff
@@ -230,7 +230,7 @@ async def test_hint_recv_timeout_seconds(
     sockets = fake_socket_pair(address,
                                packer_cls=AsyncBolt4x3.PACKER_CLS,
                                unpacker_cls=AsyncBolt4x3.UNPACKER_CLS)
-    sockets.client.settimeout = mocker.AsyncMock()
+    sockets.client.settimeout = mocker.Mock()
```
fbiville, line 233: shouldn't it remain an `AsyncMock` since we're in an async unit test?

tests/unit/async_/test_bookmark_manager.py
```diff
@@ -0,0 +1,295 @@
...
+
+@pytest.mark.parametrize("db", ("foobar", "system"))
+@mark_async_test
+async def test_return_empty_if_db_doesnt_exists(db) -> None:
+    bmm = neo4j.AsyncGraphDatabase.bookmark_manager()
+
+    assert set(await bmm.get_bookmarks(db)) == set()
+
+
+@pytest.mark.parametrize("db", ("db1", "db2", "db3"))
```
fbiville, line 53: magic

tests/unit/async_/test_driver.py
```diff
@@ -136,7 +140,7 @@ async def test_routing_driver_constructor(protocol, host, port, params, auth_tok
             ConfigurationError, "The config settings"
         ),
         (
-            {"ssl_context": ssl.SSLContext(ssl.PROTOCOL_TLSv1)},
+            {"ssl_context": ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)},
```
fbiville: this seems unrelated?

tests/unit/async_/test_driver.py
```diff
@@ -184,36 +190,38 @@ def test_driver_trust_config_error(
...
     # we set a specific db, because else the driver would try to fetch a RT
     # to get hold of the actual home database (which won't work in this
     # unittest)
+    driver._pool = fake_pool
     async with driver.session(database="foobar") as session:
-        acquire_mock = mocker.patch.object(session._pool, "acquire",
-                                           autospec=True)
-        tx_begin_mock = mocker.patch.object(AsyncTransaction, "_begin",
-                                            autospec=True)
+        # acquire_mock = mocker.patch.object(session._pool, "acquire",
```
fbiville: was ist los hier?

tests/unit/async_/test_driver.py
```diff
@@ -281,11 +291,161 @@ async def test_verify_connectivity_parameters_are_deprecated(
...
 
     try:
         with pytest.warns(ExperimentalWarning, match="configuration"):
             await driver.get_server_info(**kwargs)
     finally:
         await driver.close()
+
+
+@mark_async_test
+async def test_with_default_bookmark_manager(mocker) -> None:
```
fbiville:
```suggestion
async def test_with_builtin_bookmark_manager(mocker) -> None:
```

docs/source/async_api.rst
```diff
@@ -149,8 +151,13 @@ Async Driver Configuration
...
 
 :class:`neo4j.AsyncDriver` is configured exactly like :class:`neo4j.Driver`
-(see :ref:`driver-configuration-ref`). The only difference is that the async
-driver accepts an async custom resolver function:
+(see :ref:`driver-configuration-ref`). The only differences are:
+* the async driver accepts an async custom resolver function,
+see :ref:`async-resolver-ref`.
+* the async driver accepts accepts either a :class:`neo4j.api.BookmarkManager`
+object or a :class:`neo4j.api.AsyncBookmarkManager` as bookmark manager.
+see :ref:`async-bookmark-manager-ref`.
```
bigmontz: Nice touch

neo4j/_sync/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+                  for k, v in initial_bookmarks.items())
+        )
+        self._lock = CooperativeLock()
+
+    def update_bookmarks(
+        self, database: str, previous_bookmarks: t.Iterable[str],
+        new_bookmarks: t.Iterable[str]
+    ) -> None:
+        new_bms = set(new_bookmarks)
+        prev_bms = set(previous_bookmarks)
```
bigmontz: Do you need to create new sets here? We may let the methods `difference_update ` and `update` handle it.

neo4j/_async/work/workspace.py
```diff
@@ -77,6 +82,76 @@ def _set_cached_database(self, database):
...
+
+    async def _get_all_bookmarks(self):
+        if self._bookmark_manager is not None:
+            self._bookmarks = tuple({
+                *await AsyncUtil.callback(
+                    self._bookmark_manager.get_all_bookmarks,
+                ),
+                *self._initial_bookmarks
+            })
+        return self._bookmarks
```
bigmontz: Changing the value of `self._bookmarks` breaks the `session.last_bookmarks` contract. For instance, if you run a read query and then try to get the `session.last_bookmarks`, you will get all the bookmarks in the `BookmarkManager` instead of the bookmarks from the session.

bigmontz (changes requested):
The changes on the workspace might broke `session.last_bookmarks()` current behaviour.

bin/make-unasync
robsdedude: Have a look here https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/bin/make-unasync#L186

I patched the unasync lib to also unasync strings. This is necessary to be able to also transpile doc strings. But I made the mistake to call `_unasync_prefix` which turns classes like `class AsyncFoo` into `class Foo` and similar, but does not take extra configured replacements into consideration. I had to fix it in this PR to make lines this [this one](https://github.com/neo4j/neo4j-python-driver/pull/771/files#diff-8511a91786723e653bbd0c1068ba6f888dc44b7c976ff9f9abe96005984674f9R214) transpile correctly.

docs/source/api.rst
robsdedude: :yum: 

Do you have a suggestion for a better formulation that doesn't just sound... unnecessarily technically correct?

docs/source/api.rst
robsdedude: That tells the docs tool (sphinx) to not include the signature of the class. That's because the user is not supposed to instantiate this class themselves. Therefore, I don't want to document the constructor.

neo4j/_async/bookmark_manager.py
robsdedude: Exactly. The `None` db means "give me all the bookmarks you have". This alternative implementation is also mentioned in the [ADR (internal link)](https://github.com/neo-technology/drivers-adr/blob/12c201dfef753728a74305bd46e999fcd72c355e/adrs/001-bookmark-manager.md?plain=1#L208).

neo4j/_async/bookmark_manager.py
robsdedude: You'd think so :D I thought I'd give users the option to pass me either an async supplier or a sync one, should they not need any IO. See 

https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/docs/source/async_api.rst?plain=1#L157-L158

you can also configure the async driver with an `AsyncBMM` or a `BMM`. I also went down this route with the custom resolver function: it *can* be async (when using the async driver), but doesn't have to.

neo4j/_async/bookmark_manager.py
robsdedude: You should do this. Either in the BMM or in the session.

https://github.com/neo-technology/drivers-adr/blob/fd77505fed6eba5d3f7f5123588619fdc780a101/adrs/001-bookmark-manager.md?plain=1#L294

neo4j/_async/bookmark_manager.py
robsdedude: Yes, a simple plain lock<sup>*)<sup>.

<sup>*)</sup> The lock is actually a noop lock in the async implementation, because there is not `await` while the lock is acquired: this means there is no possibility of concurrency given that Python's async implements cooperative concurrency. I.e., you never get interrupted, unless you yield to the event loop (some `await` calls).

Furthermore, there is no builtin rwlock in Python. So I would've had to roll my own solution. That definitely seems overkill for the BMM.

neo4j/_async/bookmark_manager.py
robsdedude: Yes, that's what it does. But the set implementation in Python [1] generally uses terminology closer to mathematical sets (difference, union, ...).

1: https://docs.python.org/3/library/stdtypes.html#set

neo4j/_async/bookmark_manager.py
robsdedude: Yes, it would too. Python sets are mutable objects. But https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/neo4j/_async/bookmark_manager.py#L85 is exacly that: a copy constructor, as you'd call it in other languages. That's also why this it the only instruction being guarded by the lock.

neo4j/_async/driver.py
robsdedude: You *can* make the distinction, you don't have to. But since the driver already had these config classes about that parse kwarg dicts, I figured the least intrusive thing to do would be to only explicitly state the available config options to the type checker (most IDEs should also be able to pick this up for auto completion), but continue using `**kwargs` and the config classes at runtime.

neo4j/_async/driver.py
robsdedude: This is the config option (i.e., `neo4j.AsyncGraphDatabase.driver(..., bookmark_manager=...)`). This can only go here or at the session level and the ADR specs it to go here.

But you are still right, that it's `neo4j.AsyncGraphDatabase.bookmark_manager(**some_config)` and `neo4j.GraphDatabase.bookmark_manager(...)` whereas it's `neo4j.bookmark_manager` in JS and Go, and porbably others. But since this methods needs to be able to return an async or sync BMM, depending on what driver you want to configure, I can't put it at the top level, unless I want to go `neo4j.bookmark_manager(...)` and `neo4j.async_bookmark_manager(...)`. Which would've been another valid option, I suppose. I'm indecisive as of which option would be better. My initial thoughts were to put it next to the driver factory method as it's functionality is pretty much coupled to the driver.

neo4j/_async_compat/concurrency.py
robsdedude: `AsyncCooperativeRLock` (and with it `CooperativeRLock`) are actually being used in the Pool implementation
https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/neo4j/_async/io/_pool.py#L81
https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/neo4j/_sync/io/_pool.py#L81

As is the `(Async)RLock`
https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/neo4j/_async/io/_pool.py#L453
https://github.com/neo4j/neo4j-python-driver/blob/bd5e6a97cd3737f6c4cbe1503d3cc5ccc50fa0ab/neo4j/_sync/io/_pool.py#L453

tests/unit/async_/io/test_class_bolt4x3.py
robsdedude: Nope. That's because the method is a sync one (non-blocking) even for our async socket wrapper. Here's the definition: https://github.com/neo4j/neo4j-python-driver/blob/cc76ac5a2bb6c5933e757adee5f2415a831c78ff/neo4j/_async_compat/network/_bolt_socket.py#L143

tests/unit/async_/test_bookmark_manager.py
robsdedude: pytest's parametrization feature is incredible mighty :clap:

tests/unit/async_/test_driver.py
robsdedude: Caught :see_no_evil: It gets rid of a deprecation warning https://docs.python.org/3/library/ssl.html#ssl.PROTOCOL_TLSv1
I'll put it into it's own PR.

robsdedude: https://github.com/neo4j/neo4j-python-driver/pull/780

bin/make-unasync
robsdedude: See https://github.com/neo4j/neo4j-python-driver/pull/771#discussion_r951319852

docs/source/api.rst
robsdedude: Because there also is ``:class:`neo4j.AysncDriver` ``.

docs/source/async_api.rst
robsdedude: Thanks :bow:

neo4j/_sync/bookmark_manager.py
robsdedude: The `Iterable` type doesn't provide a `__bool__()` method. So I can't check if its empty or not without turning it into a `set`, `list`, `tuple`, or such first. So at least `new_bms` needs the conversion. `prev_bms` might not need it. But in Python there is something called generators. It's an iterable that, when consumed, can't be consumed again. This wouldn't break here, as we're only using `prev_bms` once in this method. But I felt like going the safe route.

Alternatively, I suppose, I could upgrade the required type to [`t.Collection`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Collection). That's allow me to check if the length is 0.

neo4j/_sync/bookmark_manager.py
```diff
@@ -0,0 +1,108 @@
...
+        with self._lock:
+            if not new_bms:
+                return
```
robsdedude: No need to acquire the lock for this :sweat_smile:

neo4j/_async/work/workspace.py
robsdedude: :scream:

docs/source/api.rst
fbiville: maybe `causally consistent at best`? :D
it's a nit really, feel free to ignore it

neo4j/_async/bookmark_manager.py
fbiville: I see, you want to be friends with users, don't you? ;D

neo4j/_async/bookmark_manager.py
fbiville: ah, missed that, thanks!

neo4j/_async/bookmark_manager.py
fbiville: https://github.com/neo4j/neo4j-go-driver/pull/395 🥳

docs/source/api.rst
bigmontz: The API seems to be documented using `sync` driver in most of time. For example, the `neo4j.Driver` is referred at line 145.

bigmontz (approved):

neo4j/_async/driver.py
```diff
@@ -225,19 +229,32 @@ def bookmark_manager(
     ) -> AsyncBookmarkManager:
         """Create a :class:`.AsyncBookmarkManager` with default implementation.
 
-        Basic usage example to configure the driver with the default
-        bookmark manager implementation so that all work is automatically
-        causally chained (i.e., all reads can observe all previous writes
-        even in a clustered setup)::
+        Basic usage example to configure sessions with the builtin bookmark
```
fbiville:
```suggestion
        Basic usage example to configure sessions with the built-in bookmark
```

neo4j/_async/driver.py
```diff
@@ -225,19 +229,32 @@ def bookmark_manager(
...
+        manager implementation so that all work is automatically causally
+        chained (i.e., all reads can observe all previous writes even in a
```
fbiville, line 230: causal consistency makes only sense in the context of clusters anyway, or?

neo4j/_sync/driver.py
```diff
@@ -224,19 +228,32 @@ def bookmark_manager(
     ) -> BookmarkManager:
         """Create a :class:`.BookmarkManager` with default implementation.
 
-        Basic usage example to configure the driver with the default
-        bookmark manager implementation so that all work is automatically
-        causally chained (i.e., all reads can observe all previous writes
-        even in a clustered setup)::
+        Basic usage example to configure sessions with the builtin bookmark
```
fbiville:
```suggestion
        Basic usage example to configure sessions with the built-in bookmark
```

fbiville (approved):

neo4j/_async/driver.py
robsdedude: Yes. That's why I say "even in" because you'll get this anyway if you're not working with a cluster. What do you suggest instead?

fbiville: I'd simply remove the "even" bit. Afaik, consistency models do not apply to single instances (they're supposed to be consistent with themselves else it's just... broken)

fbiville: misread the sentence, sorry

bigmontz (approved):
🔖 🧑‍💼

robsdedude:
Waiting for https://github.com/neo4j-drivers/testkit/pull/509 to be merged. Then re-run CI.

Issue #775, opened 2022-08-15
What is the tx parameter?


steve-solun:
I am trying to understand what is tx in the following code. I don't see any declaration of `tx` and will be happy to understand what is it.

```
from neo4j import GraphDatabase

driver = GraphDatabase.driver("neo4j://localhost:7687", auth=("neo4j", "password"))

def add_friend(tx, name, friend_name):
    tx.run("MERGE (a:Person {name: $name}) "
           "MERGE (a)-[:KNOWS]->(friend:Person {name: $friend_name})",
           name=name, friend_name=friend_name)

def print_friends(tx, name):
    for record in tx.run("MATCH (a:Person)-[:KNOWS]->(friend) WHERE a.name = $name "
                         "RETURN friend.name ORDER BY friend.name", name=name):
        print(record["friend.name"])

with driver.session() as session:
    session.write_transaction(add_friend, "Arthur", "Guinevere")
    session.write_transaction(add_friend, "Arthur", "Lancelot")
    session.write_transaction(add_friend, "Arthur", "Merlin")
    session.read_transaction(print_friends, "Arthur")

driver.close()
```

steve-solun:
@srbaker @mneedham @boggle @akollegger @robsdedude

steve-solun:
https://github.com/neo4j/neo4j-python-driver/blob/5.0/neo4j/_sync/work/session.py

robsdedude:
Thanks for contacting us. The docs are indeed not very explicit about this. This has already been improved upon in 5.0 https://github.com/neo4j/neo4j-python-driver/blob/799720d58a167423c35bafc1aa708b2b84a368a0/docs/source/api.rst?plain=1#L826

steve-solun:
Thanks a lot dear @robsdedude

fbiville:
In the future, please note that https://community.neo4j.com/ is a better fit for this kind of questions.

PR #778, opened 2022-08-18
Improve doc string around tx functions

robsdedude:
`session.read_transaction` and `session.write_transaction` were slightly wrong
about the tx class passed to the tx function. Furthermore, they are now more
explicit about the idempotency required from tx funcs.

Inspired by https://github.com/neo4j/neo4j-python-driver/issues/775

### DIFF
--- a/neo4j/_async/work/session.py
+++ b/neo4j/_async/work/session.py
@@ -547,7 +547,8 @@ class AsyncSession(AsyncWorkspace):
         returns, unless an exception is thrown during query execution or by
         the user code. Note, that this function performs retries and that the
         supplied `transaction_function` might get invoked more than once.
-        Therefore, it needs to be idempotent (i.e., have no side effects).
+        Therefore, it needs to be idempotent (i.e., have the same effect,
+        regardless if called once or many times).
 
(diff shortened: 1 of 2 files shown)


neo4j/_async/work/session.py
```diff
@@ -543,11 +543,11 @@ async def read_transaction(
...
-        This transaction will automatically be committed unless an exception is thrown during query execution or by the user code.
-        Note, that this function perform retries and that the supplied `transaction_function` might get invoked more than once.
-
-        Managed transactions should not generally be explicitly committed
-        (via ``await tx.commit()``).
+        This transaction will automatically be committed when the function
+        returns, unless an exception is thrown during query execution or by
+        the user code. Note, that this function performs retries and that the
+        supplied `transaction_function` might get invoked more than once.
+        Therefore, it needs to be idempotent (i.e., have no side effects).
```
fbiville: pedantic comment: it _can_ have side effects (like logging e.g.)
it's rather: it should always return the same result no matter how many times it's invoked

fbiville (approved):

PR #779, opened 2022-08-18
Fix unit tests for temporal type serialization

robsdedude:


tests/unit/common/codec/hydration/v2/test_temporal_dehydration.py
```diff
@@ -60,15 +60,15 @@ def test_native_date_time_negative_offset(self, hydration_scope):
         assert struct == Structure(b"I", 1539347861, 474716000, -3600)
 
     def test_date_time_zone_id(self, hydration_scope):
-        dt = DateTime(2018, 10, 12, 11, 37, 41, 474716862,
-                      pytz.timezone("Europe/Stockholm"))
+        dt = DateTime(2018, 10, 12, 11, 37, 41, 474716862)
```
fbiville, line 63: nit: it would be nice to document the UTC offset (especially since the date is is October, the month of the daylight shift)

fbiville (approved):

PR #784, opened 2022-08-30
Rename Session.x_transaction to Session.execute_x

robsdedude:
Deprecate `Session.read_transaction` and `Session.write_transaction` in favor
of `Session.execute_read` and `Session.execute_write` respectively.

Other drivers had to rename this function as there was a breaking change in the
API. Namely, transaction functions getting passed a managed transaction object
that has not commit/rollback/... methods, but can only run queries. The same
change also applies to the Python driver but it's not breaking as the driver
would've failed later (with a rather cryptic error coming from the server) when
the user tried to manually commit/rollback/... a managed transaction inside a
transaction function.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -128,6 +128,8 @@
   Server-side keep-alives communicated through configuration hints together with
   the config option `connection_acquisition_timeout` are sufficient to avoid the
   driver getting stuck.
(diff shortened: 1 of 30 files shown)


fbiville (approved):

gjmwoods (approved):
🚀

PR #786, opened 2022-08-31
Fix missed renamings from x_transaction to execute_x

robsdedude:
Amends https://github.com/neo4j/neo4j-python-driver/pull/784
There were other mentions/usages of `Session.read/write_transaction` that needed
replacing with `Session.execute_read/write`.

While there was a test to make sure the example in the README works, there was
no assertion the test actually tests what's in the README. Added a test to
assert the former test resembles the README example.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -31,7 +31,8 @@ Quick Example
 
     from neo4j import GraphDatabase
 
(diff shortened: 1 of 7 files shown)


fbiville (approved):

AndyHeap-NeoTech (approved):
All makes sense.

PR #789, opened 2022-09-01
[4.4] Warnings overhaul

robsdedude:
* Fix falsely emitted `DeprecationWarning` regarding
  `update_routing_table_timeout`.
* Treat warnings as errors during testing
* Changed tests to expect and catch warnings or to avoid them.

PR #790, opened 2022-09-01
Warnings overhaul

robsdedude:
* Fix potential to falsely emitted `DeprecationWarning` for deprecated
  `PoolConfig` options.
* Treat warnings as errors during testing
* Change TestKit backend to expect and catch warnings

port of https://github.com/neo4j/neo4j-python-driver/pull/789

### DIFF
--- a/neo4j/_async/driver.py
+++ b/neo4j/_async/driver.py
@@ -675,6 +675,5 @@ class AsyncNeo4jDriver(_Routing, AsyncDriver):
 
         def session(self, **config) -> AsyncSession:
             session_config = SessionConfig(self._default_workspace_config,
-                                           config)
-            SessionConfig.consume(config)  # Consume the config
+                                           SessionConfig.consume(config))
             return AsyncSession(self._pool, session_config)
--- a/neo4j/_async/driver.py
+++ b/neo4j/_async/driver.py
@@ -644,7 +644,6 @@ class AsyncBoltDriver(_Direct, AsyncDriver):
             """
             session_config = SessionConfig(self._default_workspace_config,
                                            config)
-            SessionConfig.consume(config)  # Consume the config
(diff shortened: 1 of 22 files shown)


neo4j/_async/driver.py
```diff
@@ -644,7 +644,6 @@ def session(self, **config) -> AsyncSession:
             """
             session_config = SessionConfig(self._default_workspace_config,
                                            config)
-            SessionConfig.consume(config)  # Consume the config
```
bigmontz, line 647: Why remove the `SessionConfig.consume` call?

neo4j/_async/io/_bolt.py
```diff
@@ -259,19 +259,20 @@ def get_handshake(cls):
         return b"".join(version.to_bytes() for version in offered_versions).ljust(16, b"\x00")
 
     @classmethod
-    async def ping(cls, address, *, timeout=None, **config):
+    async def ping(cls, address, *, timeout=None, pool_config=None):
```
bigmontz, line 262: You had changed the method signature, but I didn't see any change in methods which calls this one. Is it correct?

neo4j/_async/driver.py
robsdedude: Because the config options already get consolidated into a config object in the line above. This line caused warnings of deprecated options to be emitted a second time.

neo4j/_async/io/_bolt.py
robsdedude: Yes. No called made use of these config options.

bigmontz (approved):
⚠️

robsdedude:
Depends on https://github.com/neo4j-drivers/testkit/pull/513

gjmwoods (approved):

PR #801, opened 2022-09-12
Python 3.8+ handle cancellation during `HELLO`

robsdedude:
Turns out `asyncio.CancelledError` inherits from `BaseException` starting with Python 3.8.


bigmontz (approved):
😵‍💫

PR #802, opened 2022-09-12
[4.4] Fix exception handling in `Session.close()`

robsdedude:
Exceptions raised during some cleanup steps in `session.close` could have left the session in an unclean state. In particular, the session wouldn't be marked as closed. Moreover, errors encountered when consuming outstanding auto-commit results during session closure were silently swallowed.

Back port of https://github.com/neo4j/neo4j-python-driver/pull/800

### DIFF
--- a/neo4j/work/simple.py
+++ b/neo4j/work/simple.py
@@ -135,7 +135,7 @@ class Session(Workspace):
         try:
             try:
                 # consume outstanding auto-commit result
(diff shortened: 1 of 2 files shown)


robsdedude:
I don't think this work will be completed. It's a breaking change that needs coordination between drivers. Maybe we'll come back to it should it every become an actual problem.

PR #806, opened 2022-09-16
Improve Neo4jError representation

robsdedude:
When the error is not received from the DBMS, but instead originates from somewhere in the driver, it might not have a code and a message. In that case, we fall back to the default Exception representation.

Related:
 * https://github.com/neo4j/neo4j-python-driver/issues/796

### DIFF
--- a/neo4j/exceptions.py
+++ b/neo4j/exceptions.py
@@ -247,15 +247,17 @@ class Neo4jError(Exception):
         return False
 
     def __str__(self):
-        return "{{code: {code}}} {{message: {message}}}".format(code=self.code, message=self.message)
+        if self.code and self.message:
+            return "{{code: {code}}} {{message: {message}}}".format(
+                code=self.code, message=self.message
+            )
+        return super().__str__()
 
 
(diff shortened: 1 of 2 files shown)


neo4j/exceptions.py
```diff
@@ -247,15 +247,17 @@ def is_fatal_during_discovery(self) -> bool:
         return False
 
     def __str__(self):
-        return "{{code: {code}}} {{message: {message}}}".format(code=self.code, message=self.message)
+        if self.code and self.message:
+            return "{{code: {code}}} {{message: {message}}}".format(
+                code=self.code, message=self.message
+            )
+        return super().__str__()
```
bigmontz: Is it possible Neo4jErro be created only with message or code?

robsdedude: Shouldn't happen if the error comes from the server. In that case it goes through the `Neo4jError.hydrate` method, which makes sure both are set to some string. But this is still a valid concern for future-proofing the code.

bigmontz (approved):
3️⃣

PR #812, opened 2022-09-27
README: update version support

robsdedude:
* Adopt new version support policy (starting with 5.0)
 * Add version support matrix

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,7 +4,7 @@ Neo4j Bolt Driver for Python
 
 This repository contains the official Neo4j driver for Python.
 Each driver release (from 5.0 upwards) is built specifically to work with a
-corresponding Neo4j release series, i.e., that with the same ``major`` version
+corresponding Neo4j release, i.e. that with the same ``major.minor`` version
 number.
 Only the latest ``major.minor`` release of each ``major`` driver series
 receives patches and support.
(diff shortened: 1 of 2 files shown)


README.rst
```diff
@@ -103,6 +113,33 @@ Using the Python Driver 1.7 and connecting to Neo4j 4.x with default connection
...
++------------------+-------+-------+-------+-------+-------+-------+-------+
+| Neo4j 4.2 (EOL)  |   ❔   |   ❔   |   ✔️  |   ✔️  |   ✔️  |   ✔️  |   ❔   |
++------------------+-------+-------+-------+-------+-------+-------+-------+
+| Neo4j 4.3        |   ❔   |   ❔   |   ❔   |   ✔️  |   ✔️  |   ✔️  |   ❔   |
++------------------+-------+-------+-------+-------+-------+-------+-------+
+| Neo4j 4.4        |   ❔   |   ❔   |   ❔   |   ❔   |   ✔️  |   ✔️  |   ✔️  |
++------------------+-------+-------+-------+-------+-------+-------+-------+
+| Neo4j 5.0        |   ❔   |   ❔   |   ❔   |   ❔   |   ❔   |   ✔️  |   ✔️  |
++------------------+-------+-------+-------+-------+-------+-------+-------+
+
```
bigmontz: nitpick: Table not well aligned.

bigmontz (approved):
📄

README.rst
robsdedude: It is depending on if your mono font renders  `✔️` and `❔` as mono space characters. I think if it wasn't aligned, it wouldn't even be valid RST. Previously, I tried to work with `Y` and `N` instead because pretty much every pc should have a mono font that supports those, but it made it really difficult to see quickly what is and isn't supported.

robsdedude: Found another way. Using `Yes` (supported), ` ` (blank) (not working), and `?` (maybe working), gives enough visual distinctiveness imo.

AndyHeap-NeoTech (approved):
Approved.

PR #821, opened 2022-10-07
API Docs: remove double paragraph in class `neo4j.Record` section.

stefano-ottolenghi:
I favored the docstring. The content is exactly the same, so currently it shows up twice.


robsdedude (approved):
:broom:

PR #822, opened 2022-10-07
API Docs: fix missing spaces before double backticks.

stefano-ottolenghi:
Was breaking some inline code formatting.


robsdedude (approved):
:straight_ruler:

PR #830, opened 2022-10-12
4.4 for py310

Prychantovskyi:


CHANGELOG.md
```diff
@@ -30,6 +30,8 @@
 - Connections are now unencrypted by default; to reproduce former behaviour, add `encrypted=True` to Driver configuration
 - Removed `transaction.success` flag usage pattern.
 
++ Python 3.10 supported.
++ Python 3.9 supported.
```
robsdedude: This is not correct. You are targeting 4.4 and not 4.0.

Also, I would write something like

`- Python 3.10 support added (driver version 4.4.9)`

docs/source/index.rst
```diff
@@ -12,6 +12,7 @@ Neo4j versions supported:
 
 Python versions supported:
 
+* Python 3.10
```
robsdedude:
```suggestion
* Python 3.10 (added in driver version 4.4.9)
```

neo4j/conf.py
```diff
@@ -23,7 +23,10 @@
 from collections.abc import Mapping
 from warnings import warn
 
-from neo4j.meta import get_user_agent
+from neo4j.meta import (
+    deprecation_warn,
```
robsdedude:
```suggestion
```
Unused import

neo4j/conf.py
```diff
@@ -242,6 +247,10 @@ def get_ssl_context(self):
         # https://docs.python.org/3.6/library/ssl.html#ssl.PROTOCOL_TLS_CLIENT
         ssl_context = ssl.SSLContext(ssl.PROTOCOL_TLS_CLIENT)
 
+        # For recommended security options see
+        # https://docs.python.org/3.10/library/ssl.html#protocol-versions
+        ssl_context.minimum_version = ssl.TLSVersion.TLSv1_2
```
robsdedude: If you have a look at https://docs.python.org/3.10/library/ssl.html#ssl.SSLContext.minimum_version you'll find that `SSLContect.minimum_version` is 

> New in version 3.7.

Since the 4.4 driver must keep support for Python 3.6, we can't use this here or must use a condition (`if`) to find a solution pre Python 3.7 and post Python 3.7

robsdedude (changes requested):

PR #833, opened 2022-10-17
Introduce driver.execute_query

robsdedude:
Depending on 
 * https://github.com/neo4j-drivers/testkit/pull/531
 * https://github.com/neo4j-drivers/testkit/pull/540


docs/source/api.rst
```diff
@@ -151,7 +151,152 @@ Closing a driver will immediately shut down all connections in the pool.
     query, use :meth:`neo4j.Driver.verify_connectivity`.
 
 .. autoclass:: neo4j.Driver()
-   :members: session, encrypted, close, verify_connectivity, get_server_info
+    :members: session, query_bookmark_manager, encrypted, close,
+              verify_connectivity, get_server_info
+
+    .. method:: execute_query(query, parameters=None,routing=neo4j.RoutingControl.WRITERS, database=None, impersonated_user=None, bookmark_manager=self.query_bookmark_manager, result_transformer=Result.to_eager_result, **kwargs)
+
+        Execute a query inside a retired transaction and return all results.
```
bigmontz: I know it is quite difficult to explain it, but the transaction is not retired when the query is being executed.

neo4j/_async/driver.py
```diff
@@ -457,6 +485,249 @@ async def close(self) -> None:
...
+        self,
+        query: str,
+        parameters: t.Dict[str, t.Any] = None,
+        routing: T_RoutingControl = RoutingControl.WRITERS,
+        database: str = None,
+        impersonated_user: str = None,
+        bookmark_manager: t.Union[
+            AsyncBookmarkManager, BookmarkManager, None,
+            te.Literal[_DefaultEnum.default]
+        ] = _default,
```
bigmontz: Why do not use the `...` instead of the `_default`?

neo4j/_async/driver.py
```diff
@@ -457,6 +485,249 @@ async def close(self) -> None:
...
+
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore",
+                                    message=r".*\bbookmark_manager\b.*",
+                                    category=ExperimentalWarning)
+            session = self.session(database=database,
+                                   impersonated_user=impersonated_user,
+                                   bookmark_manager=bookmark_manager)
+        async with session:
+            # TODO: test all examples in the docstring
```
bigmontz: Has this todo been addressed?

neo4j/_async/work/result.py
```diff
@@ -581,6 +585,28 @@ async def data(self, *keys: _T_ResultKey) -> t.List[t.Any]:
         """
         return [record.data(*keys) async for record in self]
 
+    async def to_eager_result(self) -> EagerResult:
```
bigmontz: I might be an approach useful for other drivers. I don't know if will be that much for JS, since you can get `records` and `summary` using the statement `const { records,  summary } = await session.run(...)`. However, it could be interesting for drivers which don't have a way of getting the full records list and the summary in one single call.

bigmontz (approved):
🥇 

(Just minor comments)

neo4j/_async/driver.py
robsdedude: I actually did this in my first design but there is no typing support for such a function signature. https://github.com/python/mypy/issues/7818 https://github.com/python/typing/issues/684 So I came up with this (not so nice) workaround.

neo4j/_async/driver.py
robsdedude: Nope. Good spot, thanks!

docs/source/api.rst
robsdedude: Style guide for python docstrings demands the first paragraph (summary) to be one line. I guess I could change it to `managed transaction` and do some further explaining in the body of the docstring.

docs/source/api.rst
```diff
@@ -151,7 +151,165 @@ Closing a driver will immediately shut down all connections in the pool.
...
+            with a single underscore. This is to avoid collisions with the
+            keyword configuration parameters of this method. If you need to
+            pass such a parameter, use the ``parameters_`` parameter instead.
+            These take precedence over parameters passed as ``parameters_``.
+        :type kwargs: typing.Any
+
+        :returns: the result of the ``result_transformer``
+        :rtype: T
+
+        .. versionadded:: 5.2
```
robsdedude: :warning: before merging: make sure to adjust all of these!

docs/source/api.rst
```diff
@@ -151,7 +151,165 @@ Closing a driver will immediately shut down all connections in the pool.
     query, use :meth:`neo4j.Driver.verify_connectivity`.
 
 .. autoclass:: neo4j.Driver()
-   :members: session, encrypted, close, verify_connectivity, get_server_info
+    :members: session, query_bookmark_manager, encrypted, close,
+              verify_connectivity, get_server_info
+
+    .. method:: execute_query(query, parameters=None,routing=neo4j.RoutingControl.WRITERS, database=None, impersonated_user=None, bookmark_manager=self.query_bookmark_manager, result_transformer=Result.to_eager_result, **kwargs)
```
stefano-ottolenghi: Can we have a `:raises:` field? Or does it raise too many different exceptions?

robsdedude: There are plenty of errors this can raise. Here are the one's I can think of on the spot
 * `Neo4jError` (or any sub-class) - anything the server complains about
 * `SessionExpired`, `ServiceUnavailable` (or any sub-class) - lost/couldn't establish connectivity (after the retries this API includes)
 * `ConfigurationError` `TypeError` `ValueError` (maybe even `IndexError`?) - invalid configuration/parameters or trying to use a feature that's too new for the server you're connected to
 * `BoltError` (or any sub-class; internal error class you shouldn't ever encounter) - the server violated the protocol
 
 There might be more that I can't think of right now. My conclusion:
 * All of there errors are fatal (there is pretty much no benefit in giving them different treatment). The user might as well just catch any exception, maybe log it or whatever they feel doing, and then give up (that could mean crash, surface some error to the end-user, whatever is suitable). Plus maybe fixing their code afterwards.
 * I can hardly give any guarantees what errors this method can raise as it pretty much exercises the whole stack of the driver. It's highly likely that I'll forget some corner of the driver and, even if not, that this piece of documentation will unsync sooner or later as it's so far up the abstraction stack.
 
 So much for why I didn't include it. If you think it'd be helpful nonetheless, we can make it happen.

stefano-ottolenghi: > The user might as well just catch any exception

:+1:

bigmontz (approved):
🌵 🌵 🌵

PR #836, opened 2022-10-18
Improve keyword query parameters

robsdedude:
* API docs: clarify precedence between keyword and dict query parameters
 * Fix: parameters of internal functions could clash with keyword parameters. E.g., `db` and `imp_user` could not be used as keyword query parameters causing a not necessarily easy to understand error message. This got fixed by merging keyword and dict parameters at the top-most level.

Backport of https://github.com/neo4j/neo4j-python-driver/pull/835

### DIFF
--- a/neo4j/work/result.py
+++ b/neo4j/work/result.py
@@ -63,21 +63,19 @@ class Result:
         else:
             return self._raw_qid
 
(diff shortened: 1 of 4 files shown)


bigmontz (approved):
🍎

(is there any test for checking the precedence?)

PR #837, opened 2022-10-20
Fix memoryviews not being closed

robsdedude:
In rare cases, `BufferError: Existing exports of data: object cannot be re-sized` could occur.


robsdedude:
Superseded by https://github.com/neo4j/neo4j-python-driver/pull/843

PR #838, opened 2022-10-21
API docs: remove options that were removed in 5.0

robsdedude:
`session_connection_timeout` and `update_routing_table_timeout` were introduced in 4.4.5, deprecated in 4.4.6, and removed in 5.0.0 as they were deemed to be superfluous. However, the 5.x docs did not reflect this.


bigmontz (approved):
😮

PR #854, opened 2022-11-08
Make code compatible with mypy 0.990

robsdedude:
* remove implicit optional parameters (PEP 484 received an update)
 * mute metaclass inference limitations of mypy
 * fix mypy rightfully complaining about empty method bodies
 * fixed bug in (deprecated) `session.last_bookmark` that mypy uncovered


neo4j/_async/work/session.py
```diff
@@ -321,7 +323,7 @@ async def last_bookmark(self) -> t.Optional[str]:
         if self._auto_result:
             await self._auto_result.consume()
 
-        if self._transaction and self._transaction._closed:
+        if self._transaction and self._transaction._closed():
```
thelonelyvulpes, line 326: this seems unrelated to mypy

bigmontz (approved):
🥇 

(Quite nice bug solved 🎁 )

neo4j/_async/work/session.py
bigmontz: the linter may point out `closed` is a function, not a property.

thelonelyvulpes (approved):

PR #856, opened 2022-11-15
Package `py.typed` marker file

robsdedude:
https://peps.python.org/pep-0561/#packaging-type-information for a little bit of context.

I just missed to bundle the marker file with the source distribution, effectively leaving the driver without typing...


thelonelyvulpes (approved):
🦆

bigmontz (approved):
⌨️

PR #859, opened 2022-11-22
Bookmark manager no longer tracks per database

robsdedude:
- The experimental bookmark manager feature was changed to no longer track bookmarks per database. This effectively changes the signature of almost all bookmark manager related methods:
  - `neo4j.BookmarkManger` and `neo4j.AsyncBookmarkManger` abstract base classes: - ``update_bookmarks`` has no longer a ``database`` argument. - ``get_bookmarks`` has no longer a ``database`` argument. - The ``get_all_bookmarks`` method was removed. - The ``forget`` method was removed.
  - `neo4j.GraphDatabase.bookmark_manager` and `neo4j.AsyncGraphDatabase.bookmark_manager` factory methods:
    - ``initial_bookmarks`` is no longer a mapping from database name to bookmarks but plain bookmarks.
    - ``bookmarks_supplier`` no longer receives the database name as an argument. - ``bookmarks_consumer`` no longer receives the database name as an argument.

Depends on
 - https://github.com/neo4j-drivers/testkit/pull/534


tests/unit/async_/test_bookmark_manager.py
```diff
@@ -42,243 +42,152 @@ def bookmark_manager(*args, **kwargs):
...
+        set(initial_bookmarks or []) - bookmarks_old | bookmarks_new
+    )
+    assert args[0].raw_values == expected_bookmarks
 
 
 @pytest.mark.parametrize("consumer_async", consumer_async_options)
 @pytest.mark.parametrize("with_initial_bookmarks", (True, False))
-@pytest.mark.parametrize("db", ("db1", "db2"))
 @mark_async_test
 async def test_does_not_notify_on_empty_new_bookmark_set(
```
bigmontz, line 160: You won't need to ensure this since the session will never call `update_bookmarks` with `new_bookmarks=[]`.

bigmontz (approved):
🔖 👔

tests/unit/async_/test_bookmark_manager.py
robsdedude: While it's true that the session wouldn't call `update_bookmarks` with empty bookmarks, nothing is stopping user code to do so. `update_bookmarks` is publicly available and documented. No reason not to call it.

PR #860, opened 2022-11-22
Code clean-up

robsdedude:
* type alias and variable naming
 * use relative imports
 * specify error causes when rewriting (`raise ... from ...`)


bigmontz (approved):
🧹

PR #867, opened 2022-11-24
Modernize packaging

robsdedude:
* Introducing `pyproject.toml` https://peps.python.org/pep-0621/
 * Migrate from flat-layout to src-layout
 * Deprecate legacy package name `neo4j-driver` in favor of `neo4j` + warn about it


bigmontz (approved):
❌

PR #874, opened 2022-12-07
API docs: elaborate address resolver docs

robsdedude:


src/neo4j/addressing.py
```diff
@@ -79,22 +79,83 @@ def ipv6_cls(self):
...
         s: str,
         default_host: t.Optional[str] = None,
         default_port: t.Optional[int] = None
-    ) -> _TAddress:
+    ) -> Address:
+        """Parse a string into an address.
+
+        The string must be in the format ``host:port`` (IPv4) or
+        ``[host]:port`` (IPv6).
+        If not port is specified, or is emtpy, ``default_port`` will be used.
```
stefano-ottolenghi: "If `no` port ... `empty`"

src/neo4j/addressing.py
```diff
@@ -79,22 +79,83 @@ def ipv6_cls(self):
...
+            >>> Address.parse("[::1]:7687")
+            IPv6Address(('::1', 7687, 0, 0))
+            >>> Address.parse("localhost")
+            IPv4Address(('localhost', 0))
+            >>> Address.parse("localhost", default_port=1234)
+            IPv4Address(('localhost', 1234))
+
+        :param s: The string to parse.
+        :param default_host: The default host to use if none is specified.
+            :data:`None` will be substituted with ``"localhost"``.
```
stefano-ottolenghi: Maybe it's just me, but `substituted` feels so latin/archaic. Why not `replaced`? Same below. Feel free to resolve in any case.

src/neo4j/addressing.py
```diff
@@ -79,22 +79,83 @@ def ipv6_cls(self):
 
 
 class Address(tuple, metaclass=_AddressMeta):
+    """Base class to represent server addresses within the driver.
+
+    A tuple of two (IPv4) or four (IPv6) elements, representing the address
+    parts. See also python's :mod:`socket` module for more information.
```
stefano-ottolenghi, line 85: Is it lengthy to explain what the full signature is? If not, I would consider adding it. I am confused by `Address(("::1", 7687, 0, 0))` as it's not immediate what the third and fourth arguments do.

docs/source/async_api.rst
```diff
@@ -163,29 +163,41 @@ For example:
...
+        yield neo4j.Address(("::1", 7687, 0, 0))  # IPv6
+        yield neo4j.Address.parse("localhost:7687")
+        yield neo4j.Address.parse("[::1]:7687")
+
+        # or any tuple that can be passed to neo4j.Address(...).
+        # Initially, this will be interpreted as IPv4, but DNS resolution
+        # will turn it into IPv6 if appropriate.
+        yield "::1", 7687
+        # This will be interpreted as IPv6 directly, but DNS resolution will
+        # happen still.
```
stefano-ottolenghi: I'd swap the order of `happen` and `still`.

docs/source/async_api.rst
```diff
@@ -163,29 +163,41 @@ For example:
...
+        yield "::1", 7687, 0, 0
```
stefano-ottolenghi, line 186: One question this example raises for me is why I should use this way, when I can just do `yield "::1", 7687` and it has the same effect? Is there any advantage in being explicit with 3rd and 4th entries?

bigmontz (approved):
📖

stefano-ottolenghi (commented):
Great that you took time to document the `Address` objects!

src/neo4j/addressing.py
robsdedude: It is rather involved depending on how deep you want to dig into the rabbit hole. The full IPv6 signature is `(host, port, flowinfo, scope_id)` (as of Python's `socket` module documentation wich I linked here). There are two things to consider here:
 1) flow info and scope id are well beyond the scope of the driver's API docs, so I def. wouldn't go deeper than that.
 2) the driver uses this 4-tuple and, for the most parts, passes it as-is into Python built-in APIs. So by keeping this generic, an admittedly unlikely change in those APIs will transparently propagate through the driver's `Address` classes.

If you think the sole signature of the tuples (in the form I wrote above) might be helpful to uses, I'm happy to include that in the docs.

docs/source/async_api.rst
robsdedude: Quoting the Python docs

> For [socket](https://docs.python.org/3/library/socket.html#module-socket) module methods, flowinfo and scope_id can be omitted just for backward compatibility. Note, however, omission of scope_id can cause problems in manipulating scoped IPv6 addresses.

So while I can't come up with an example, because frankly I don't know enough about IPv6, there might be.

src/neo4j/addressing.py
```diff
@@ -150,9 +150,9 @@ def parse(
 
         :param s: The string to parse.
         :param default_host: The default host to use if none is specified.
-            :data:`None` will be substituted with ``"localhost"``.
+            :data:`None` idicates to use ``"localhost"`` as default.
```
stefano-ottolenghi: typo -> `indicates`, same two lines below.

src/neo4j/addressing.py
stefano-ottolenghi: I hoped the signature alone to be more descriptive, but at a glance I have no idea of what `flowinfo` and `scope_id` would be, so I don't think it would be of use to 80% of other people. We can drop this.

stefano-ottolenghi (approved):

src/neo4j/addressing.py
stefano-ottolenghi: Feel free to merge after this :dango:

PR #881, opened 2022-12-30
Accept more types as query parameters

robsdedude:
Implement serialization support for
 * python types: tuple
 * pandas types:
   * all but period, interval, pyarrow
 * numpy types:
   * all but void, complexfloating

pyarrow support was not implemented as it would either require more ifs in the recursive packing function, making it (driver's hot-path) slower for non-pyarrow use-cases. Or alternatively, transformers would have to be used making pyarrow type serialization rather slow.


thelonelyvulpes:
Why don't we have a user-specified encoding used for converting python:bytes->bolt:String?

docs/source/api.rst
```diff
@@ -1110,6 +1110,57 @@ The diagram below illustrates the actual mappings between the various layers, fr
...
+| pandas ``Series``                 | List                            | :class:`list`                         |
++-----------------------------------+---------------------------------+---------------------------------------+
+| pandas ``Array``                  | List                            | :class:`list`                         |
++-----------------------------------+---------------------------------+---------------------------------------+
+
+.. Note::
+
+   2. ``void`` and ``complexfloating`` typed numpy ``ndarray``\s are not supported.
+   3. ``Period``, ``Interval``, and ``pyarrow`` pandas types are not supported.
+   4. A pandas ``DataFrame`` will be serialized Map from with the column names mapping to the column values (as lists).
```
fbiville:
```suggestion
   4. A pandas ``DataFrame`` will be a serialized Map with the column names mapped to the column values (as lists).
```

src/neo4j/_codec/hydration/v1/temporal.py
```diff
@@ -22,10 +22,28 @@
     timedelta,
 )
 
+
+try:
```
fbiville: is there a way to run this check only once or do you have to repeat it everywhere you need it?

src/neo4j/_codec/hydration/_interface/__init__.py
```diff
@@ -18,11 +18,14 @@
...
 
+from .._common import DehydrationHooks
+
 
 class HydrationHandlerABC(abc.ABC):
     def __init__(self):
         self.struct_hydration_functions = {}
-        self.dehydration_functions = {}
+        self.dehydration_hooks = DehydrationHooks(exact_types={},
+                                                  subtypes={})
```
fbiville, line 28: are you using subtypes somewhere? I've not seen it yet

fbiville (approved):

bigmontz (approved):
📦

src/neo4j/_codec/hydration/v1/temporal.py
robsdedude: I guess I could move that into a single file and import from there... Not sure I like that better than this approach. I'll try it out and see what it feels like.

src/neo4j/_codec/hydration/_interface/__init__.py
robsdedude: No, I don't. It's presence doesn't hurt performance-wise though (only if there was something in there, could the driver become noticeably slower). I thought I might need it and ended up not needing it after all. But I still decided to leave it in as it will come in handy should we at some point allow custom dehydration hooks ([internal Trello card](https://trello.com/c/NajZu4P0)).

src/neo4j/_codec/packstream/v1/__init__.py
```diff
@@ -16,16 +16,55 @@
...
+# and we don't want to treat them as sequences
+SEQUENCE_TYPES: t.Tuple[t.Type, ...] = (list,)
+MAPPING_TYPES: t.Tuple[t.Type, ...] = (dict,)
+BYTES_TYPES: t.Tuple[t.Type, ...] = (bytes, bytearray)
+
+
+try:
+    import numpy as np
+
+    TRUE_VALUES += (np.bool_(True),)
```
thelonelyvulpes: Readability wise it might be better if all of the lines used the same way of extending the list. 
i.e. `TRUE_VALUES = (*TRUE_VALUES, np.bool_(True))`

thelonelyvulpes (approved):

PR #887, opened 2023-01-05
Fix `ResourceWarning` when canceling `await driver.close()`

robsdedude:


robsdedude:
CI should turn green with https://github.com/neo4j/neo4j-python-driver/pull/886

gjmwoods (approved):

PR #890, opened 2023-01-13
Introduce auth rotation and session auth support

### LINKED ISSUE #834: [Feature Request] Ability to refresh auth token
To help us understand your issue, please specify important details, primarily:

- Neo4j version: AWS Neptune
- Neo4j Mode: N/A
- Driver version: All
- Operating system: Linux

We're using openCypher with AWS Neptune and want to use the neo4j driver with the bolt protocol. We are generating an AWS V4 signature and passing it in as a basic auth to the driver. The request is signed using temporary credentials that expires in 5 minutes.

Unfortunately, it seems that the authentication token is [cached at connection creation time](https://github.com/neo4j/neo4j-python-driver/blob/5.0/neo4j/_sync/io/_bolt.py#L146).
```
# Determine auth details
if not auth:
    self.auth_dict = {}
elif isinstance(auth, tuple) and 2 <= len(auth) <= 3:
    from neo4j import Auth
    self.auth_dict = vars(Auth("basic", *auth))
else:
    try:
        self.auth_dict = vars(auth)
    except (KeyError, TypeError):
        raise AuthError("Cannot determine auth details from %r" % auth)
```

Thus, if the `max_connection_lifetime` is set to less than 5 minutes, requests start failing after 5 minutes because the signature is no longer valid.
```
self = <neo4j.io._common.InitResponse object at 0x7f41eb9dc750>
metadata = {'code': '"BoltProtocol.unexpectedException"', 'message': '"Unexpected server exception \'Signature expired: 20221018T003343Z is now earlier than 20221018T003344Z (20221018T003844Z - 5 min.)\'"'}

    def on_failure(self, metadata):
        code = metadata.get("code")
        if code == "Neo.ClientError.Security.Unauthorized":
            raise Neo4jError.hydrate(**metadata)
        else:
            raise ServiceUnavailable(
>               metadata.get("message", "Connection initialisation failed")
            )
E           neo4j.exceptions.ServiceUnavailable: "Unexpected server exception 'Signature expired: 20221018T003343Z is now earlier than 20221018T003344Z (20221018T003844Z - 5 min.)'"
```

The [Javascript driver also has the same issue](https://github.com/neo4j/neo4j-javascript-driver/issues/993).

But the Java driver provides a way to avoid this problem by exposing a [toMap](https://github.com/neo4j/neo4j-java-driver/blob/5.0/driver/src/main/java/org/neo4j/driver/internal/security/InternalAuthToken.java#L42) method, that is called to obtain the token. While the cached token is returned by default, this method can be overridden in a subclass as [shown here](https://docs.aws.amazon.com/neptune/latest/userguide/access-graph-opencypher-bolt.html#access-graph-opencypher-bolt-java-iam-auth).

We would like the Python driver to also provide an option not to cache the auth token and regenerate it on demand.

robsdedude:
# ADR 012: re-authentication

This PR introduces two new auth mechanics for different use-cases
1) Auth Rotation
2) Session Auth (a.k.a. user switching)

**Note that all APIs introduced in this PR are previews**
See https://github.com/neo4j/neo4j-python-driver/wiki/preview-features

## 1) Auth Rotation
This is used for auth tokens that is expected to expire (e.g., SSO).

A `neo4j.auth_management.AuthManager` instance 
(or `neo4j.auth_management.AsyncAuthManager` for async driver) may be passed
to the driver instead of a static auth token.

```python
import neo4j
from neo4j.auth_management import AuthManager


class MyManager(AuthManager):
    ...  # see API dos for details


with neo4j.GraphDatabase.driver(
    "neo4j://example.com:7687", 
    auth=MyManager(),
) as driver:
    ...
```

The easiest way to get started is using the provided `AuthManager`
implementation. For example:

```python
import neo4j
from neo4j.auth_management import (
    AuthManagers,
    ExpiringAuth,
)


def auth_provider():
    # some way to getting a token
    sso_token = get_sso_token()
    # assume we know our tokens expire every 60 seconds
    expires_in = 60
    return ExpiringAuth(
        auth=neo4j.bearer_auth(sso_token),
        # Include a little buffer so that we fetch a new token
        # *before* the old one expires
        expires_in=expires_in - 10
    )


with neo4j.GraphDatabase.driver(
    "neo4j://example.com:7687",
    auth=AuthManagers.expiration_based(auth_provider)
) as driver:
    ... 
```

**Note**  
This API is explicitly *not* designed for switching users.
In fact, the token returned by each manager must always belong to the same
identity. Switching identities using the `AuthManager` is undefined behavior.

## 2) Session Auth
For the purpose of switching users, `Session`s can be configured with a static
auth token. This is very similar to impersonation in that all work in the
session will be executed in the security context of the user associated with
the auth token. The major difference is that impersonation does not require or
verify authentication information of the target user, however it requires
the impersonating user to have the permission to impersonate.

**Note**
This requires Bolt protocol version 5.3 or higher (Neo4j DBMS 5.8+).

```python
import neo4j


with neo4j.GraphDatabase.driver(
    "neo4j://example.com:7687",
    auth=("user1", "password1"),
) as driver:
    with driver.session(database="neo4j") as session:
        ...  # do some work as user1
    with driver.session(database="neo4j",
                        auth=("user2", "password2")) as session:
        ...  # do some work as user2
```

## References
Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/539

Related PRs:
 * https://github.com/neo4j/neo4j-python-driver/pull/891

Issues:
 * closes https://github.com/neo4j/neo4j-python-driver/issues/834


docs/source/api.rst
```diff
@@ -348,7 +370,7 @@ Closing a driver will immediately shut down all connections in the pool.
         :returns: the result of the ``result_transformer``
         :rtype: T
 
-        **This is experimental.** (See :ref:`filter-warnings-ref`)
+        **This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 373: Should this be preview?

docs/source/api.rst
```diff
@@ -433,7 +458,7 @@ Specify whether TCP keep-alive should be enabled.
 :Type: ``bool``
 :Default: ``True``
 
-**This is experimental.** (See :ref:`filter-warnings-ref`)
+**This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 461: Should this be preview?

docs/source/api.rst
```diff
@@ -995,10 +1022,33 @@ See :class:`.BookmarkManager` for more information.
 
 .. versionadded:: 5.0
 
-**This is experimental.** (See :ref:`filter-warnings-ref`)
+**This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 1025: Same question, should this be preview?

docs/source/api.rst
```diff
@@ -1293,7 +1343,7 @@ Graph
 
     .. automethod:: relationship_type
 
-**This is experimental.** (See :ref:`filter-warnings-ref`)
+**This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 1346: Preview?

docs/source/async_api.rst
```diff
@@ -118,6 +118,20 @@ Each supported scheme maps to a particular :class:`neo4j.AsyncDriver` subclass t
     See https://neo4j.com/docs/operations-manual/current/configuration/ports/ for Neo4j ports.
 
 
+.. _async-auth-ref:
+
+Async Auth
+==========
+
+Authentication mostly works the same as in the synchronous driver.
```
AndyHeap-NeoTech: Devs may wonder what mostly means and start asking whats the difference. If they don't need to know about it then remove the 'mostly'. If it is something to be aware of then some details might help.

docs/source/async_api.rst
```diff
@@ -329,7 +360,7 @@ Closing a driver will immediately shut down all connections in the pool.
         :returns: the result of the ``result_transformer``
         :rtype: T
 
-        **This is experimental.** (See :ref:`filter-warnings-ref`)
+        **This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 363: Preview?

docs/source/async_api.rst
```diff
@@ -622,7 +659,7 @@ See :class:`BookmarkManager` for more information.
 :Type: :data:`None`, :class:`BookmarkManager`, or :class:`AsyncBookmarkManager`
 :Default: :data:`None`
 
-**This is experimental.** (See :ref:`filter-warnings-ref`)
+**This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 662: Preview?

docs/source/api.rst
robsdedude: Nope. Since it was marked experimental before and the Python driver emits an `ExperimentalWarning` for those APIs. Changing this to `PreviewWarning` would be a breaking change. So the new wording will only be applied to new preview APIs.

docs/source/api.rst
robsdedude: See https://github.com/neo4j/neo4j-python-driver/pull/890/#discussion_r1162570383

docs/source/api.rst
robsdedude: See https://github.com/neo4j/neo4j-python-driver/pull/890/#discussion_r1162570383

docs/source/api.rst
robsdedude: See https://github.com/neo4j/neo4j-python-driver/pull/890/#discussion_r1162570383

docs/source/async_api.rst
robsdedude: See https://github.com/neo4j/neo4j-python-driver/pull/890/#discussion_r1162570383

docs/source/async_api.rst
robsdedude: See https://github.com/neo4j/neo4j-python-driver/pull/890/#discussion_r1162570383

docs/source/async_api.rst
robsdedude: Mostly refers to the next sentence. I'll make this more explicit :ok_hand:

src/neo4j/_async/work/result.py
```diff
@@ -624,7 +624,7 @@ async def to_eager_result(self) -> EagerResult:
             was obtained has been closed or the Result has been explicitly
             consumed.
 
-        **This is experimental.** (See :ref:`filter-warnings-ref`)
+        **This is experimental** (see :ref:`filter-warnings-ref`).
```
AndyHeap-NeoTech, line 627: Preview :)

docs/source/async_api.rst
robsdedude: fixed in 7065b8456b0516d1cc871aba32f9d7faa33b1f78

AndyHeap-NeoTech (approved):
All looks good, just those experimental tags that needed swapping to preview.

src/neo4j/_async/work/result.py
robsdedude: see https://github.com/neo4j/neo4j-python-driver/pull/890#discussion_r1162570383

docs/source/api.rst
AndyHeap-NeoTech: Understood.

src/neo4j/_async/work/result.py
AndyHeap-NeoTech: Understood.

PR #895, opened 2023-02-02
API docs: result transformer must be idempotent

robsdedude:
Add note to the docs that `Driver.execute_query`'s `result_transformer_` argument must be an idempotent function as the driver might retry the underlying transaction.


bigmontz (approved):
🍨

PR #905, opened 2023-03-29
Fix bolt handshake not having a timeout

robsdedude:
During the bolt version negotiation the socket was not constrained by any timeout. This could, in rare occasions, cause the driver to get stuck until some other mechanism (e.g., socket keep alive) would free it.

The handshake is now limited by the `connection_acquisition_timeout`.

Introducing tests in https://github.com/neo4j-drivers/testkit/pull/546


bigmontz (approved):
⏲️

AndyHeap-NeoTech (approved):
With my limited Python this appears to all look good to me. Was initially questioning the need for separate timeouts for TCP and handshake, but on thinking about it I believe you are correct in taking that route.

PR #906, opened 2023-03-29
Improve sync TestKit backend performance

robsdedude:
The sync version of the TestKit backend was over 4 times as slow as its async counterpart. The reason is that the TCPServer shipped with Python hands out an unbuffered writer by default.

I don't quite understand why this wasn't a problem earlier. The sync backend code did not change, neither did the TCPServer implementation in the standard library. It shall remain a mystery *insert mystical sound effect*.


AndyHeap-NeoTech (approved):
Huzzah, crack out the beer in celebration! One of the better quality of life improvement PRs I've had the pleasure of approving.

PR #908, opened 2023-03-30
Speed up integration tests

robsdedude:
Speedup is achieved by reducing the number of repetitions of randomized tests. This cuts the total time for ITs roughly in half.


thelonelyvulpes (approved):
that is less.

bigmontz (approved):
🏃🏼

PR #913, opened 2023-04-06
Fix dev dependencies

robsdedude:
The driver has support for pandas >=1.1, <2, but as optional dependencies. If not specified (`pip install neo4j[pandas]`), but installed directly (`pip install pandas`), the version bounds of the driver will not be respected.


AndyHeap-NeoTech (approved):
LGTM

PR #914, opened 2023-04-06
Support serializing datetimes with datetime.timezone

robsdedude:
This enables users to use python's standard `datetime.timezone` for constant utc-offset timezones (only for serialization, data coming back from the server will continue to use `pytz`).

This change also enables pandas 2 as pandas switched over to `datetime.timezone` for constant offset timezones
https://pandas.pydata.org/docs/dev/whatsnew/v2.0.0.html#utc-and-fixed-offset-timezones-default-to-standard-library-tzinfo-objects

### DIFF
--- a/src/neo4j/_codec/hydration/v1/temporal.py
+++ b/src/neo4j/_codec/hydration/v1/temporal.py
@@ -20,7 +20,10 @@ from datetime import (
     datetime,
     time,
     timedelta,
+    timezone,
 )
+
+import pytz
 
 from ...._optional_deps import (
     np,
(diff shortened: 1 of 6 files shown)


src/neo4j/_codec/hydration/v1/temporal.py
```diff
@@ -172,10 +173,15 @@ def seconds_and_nanoseconds(dt):
         seconds, nanoseconds = seconds_and_nanoseconds(value)
         return Structure(b"f", seconds, nanoseconds, tz.key)
     else:
+        if isinstance(tz, timezone):
+            # offset of the timezone is constant, so any date will do
```
fbiville, line 180: if any date will do, why not keeping the statement `tz.utcoffset(value)`?

fbiville (approved):

src/neo4j/_codec/hydration/v1/temporal.py
robsdedude: Because the built-in `timezone` only works with the built-in `datetime` type. But `value` might (or might not) be the driver's custom `neo4j.temporal.DateTime` type.

thelonelyvulpes (approved):

src/neo4j/_codec/hydration/v1/temporal.py
```diff
@@ -172,10 +173,15 @@ def seconds_and_nanoseconds(dt):
...
+            offset = tz.utcoffset(datetime(1970, 1, 1))
```
thelonelyvulpes: a minor (probably pointless) optimization might be defining or using a preexisiting datetime constant. if we are serializing a bunch of these with timezone it's just create a datetime each iter.

PR #917, opened 2023-04-17
Temporarily freeze pip to avoid deprecation warning

robsdedude:


robsdedude:
Underlying pip issue has been resolved. This PR isn't needed any longer.

PR #928, opened 2023-05-16
Change ExpiringAuth.expires_in to expires_at

robsdedude:
Updating the preview feature to be more in line with other drivers and support more SSO flows, of which some specify token lifetime with an absolute time value while other flows use relative time.

Now, there's both options:
 * absolute: `ExpiringAuth(some_auth, expires_at=...)`
 * relative: `ExpiringAuth(some_auth).expires_in(...)` (or doing it manually `ExpiringAuth(some_auth, expires_at=time.time() + expires_in)`)

### DIFF
--- a/src/neo4j/_auth_management.py
+++ b/src/neo4j/_auth_management.py
@@ -24,9 +24,13 @@
 import abc
 import time
 import typing as t
+import warnings
 from dataclasses import dataclass
 
-from ._meta import preview
+from ._meta import (
+    preview,
+    PreviewWarning,
+)
 from .api import _TAuth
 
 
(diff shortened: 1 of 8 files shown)


tests/unit/sync/test_auth_manager.py
```diff
@@ -82,52 +82,51 @@ def test_static_manager(
...
+        provider.reset_mock()
+        assert manager.get_auth() is auth2
+        provider.assert_not_called()
 
 
 @mark_sync_test
 @pytest.mark.parametrize(("auth1", "auth2"),
                          itertools.product(SAMPLE_AUTHS, repeat=2))
-@pytest.mark.parametrize("expires_in", (None, -1, 1., 1, 1000.))
+@pytest.mark.parametrize("expires_at", (None, -1, 1., 1, 1000.))
```
fbiville, line 119: they really made a typo in the decorator name? ouch :/

fbiville (approved):

ConorNeo (approved):

src/neo4j/_auth_management.py
```diff
@@ -52,8 +59,37 @@ class ExpiringAuth:
 
     .. versionadded:: 5.8
     """
-    auth: _TAuth
-    expires_in: t.Optional[float] = None
+    auth: "_TAuth"
+    expires_at: t.Optional[float] = None
+
+    def expires_in(self, seconds: float) -> "ExpiringAuth":
```
thelonelyvulpes, line 65: Why is the `ExpiringAuth` in double quotes?

tests/unit/sync/test_auth_manager.py
robsdedude: You mean parametrized vs. parameterized?

Seems to be a controversial topic :fire: https://english.stackexchange.com/questions/79847/parametrise-or-parameterise-a-curve

src/neo4j/_auth_management.py
robsdedude: That's to defer the evaluation of the type hint as the class is not yet defined (we're in its definition).

This could also be achieved with `from __future__ import annotations` (or by using Python 3.10+ where that's the default behavior), but we have to keep support for Python 3.7 and using the `__future__` import breaks the docs generation tool (see [line 19 of the same file](https://github.com/neo4j/neo4j-python-driver/pull/928/files#diff-afe23d99150a31c4a8b55a545528304f4e27b3915d6301b75145e2874677fbb7R19)).

**EDIT**: it was announced to become a thing in Python 3.10, but that changed, apparently.

thelonelyvulpes: Many thanks.

PR #932, opened 2023-05-25
ADR 015: home db caching

robsdedude:
Add `max_home_database_delay` config option to the driver. Defines an upper bound for how long (in seconds) a resolved home database can be cached. It defaults to `5.0` seconds.

This should improve performance by reducing the extra round trips needed for resolving the home database. However, this also means that home database changes might become visible later in the driver than they used to. There are different options how to tackle this, should this be a problem.

1. Reduce the value of `max_home_database_delay`. This can go as far to to set it to `0` effectively restoring the driver's behavior before this change. However, extrem values are likely to incur a significant performance penalty (driver and server side).
2. Call the new `driver.force_home_database_resolution()` to make the driver forget any cached home dbs. This is especially useful after a home database change has been issued and needs to be readable from the same application directly afterwards.

Depends on
 * https://github.com/neo4j-drivers/testkit/pull/569

Should be ported back to 4.4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,16 @@
 See also https://github.com/neo4j/neo4j-python-driver/wiki for a full changelog.
 
 ## NEXT RELEASE
+- To improve the performance of the driver, home database caching has been
+  introduced. Other than a speed-up, this shouldn't have an effect on most
+  use-cases.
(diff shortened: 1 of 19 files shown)


robsdedude:
ADR has been abandoned because the change was considered too risky. Better approaches are being discussed.

PR #940, opened 2023-06-28
Fix handling of sub-ms transaction timeouts

robsdedude:
Transaction timeouts are specified in seconds as float. However, the server expects it in milliseconds as int. This would lead to
 1) rounding issues: previously, the driver would multiply by 1000 and
    then truncate to int. E.g., 256.4 seconds would be turned into 256399 ms
    because of float imprecision.
    Therefore, the built-in `round` is now used instead.
 2) values below 1 ms (e.g., 0.0001) would be rounded down to 0. However, 0 is
    a special value that instructs the server to not apply any timeout. This
    is likely to surprise the user which specified a non-zero timeout. In this
    special case, the driver now rounds up to 1 ms.

### DIFF
--- a/tests/unit/async_/io/test_class_bolt3.py
+++ b/tests/unit/async_/io/test_class_bolt3.py
@@ -324,6 +324,10 @@ async def test_does_not_send_bolt_agent(fake_socket_pair, user_agent):
         (3.456, 3456),
         (1, 1000),
         (
+            -1e-15,
+            ValueError("Timeout must be a positive number or 0")
+        ),
+        (
             "foo",
             ValueError("Timeout must be specified as a number of seconds")
         ),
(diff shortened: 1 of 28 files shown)


tests/unit/async_/io/test_class_bolt3.py
```diff
@@ -297,3 +297,61 @@ async def test_does_not_send_bolt_agent(fake_socket_pair, user_agent):
...
+@mark_async_test
+@pytest.mark.parametrize(
+    ("func", "args", "extra_idx"),
+    (
+        ("run", ("RETURN 1",), 2),
+        ("begin", (), 0),
+    )
+)
+@pytest.mark.parametrize(
+    ("timeout", "res"),
```
bigmontz, line 311: Should test if negative values trigger the exception.

bigmontz (approved):
👻

PR #942, opened 2023-06-29
Avoid unclean interpreter shutdown

robsdedude:
When not closing the driver before shutting down the interpreter, the GC might call `__del__` on the driver, which currently (will be changed with 6.0) closes the driver. So any code that's run on driver closure might be execute just before the interpreter shuts down. This can lead to funky errors when trying to import things dynamically. So this should be avoided.

This got raised here:
https://community.neo4j.com/t/error-with-the-neo4j-python-driver-importerror-sys-meta-path-is-none-python-is-likely-shutting-down/63096/


AndyHeap-NeoTech (approved):
Looks to make sense to me.

PR #943, opened 2023-07-05
TestKit: skip new tests that need fixes in the driver

robsdedude:
Adjustment for new tests and assertions added in https://github.com/neo4j-drivers/testkit/pull/574


bigmontz (approved):
🛹

PR #947, opened 2023-07-17
Reduce scope of pip warnings as errors

robsdedude:
The CI has been failing too often recently because of changes in pip and
dependencies. Those failures were mostly irrelevant warning that we turned into
errors for better visibility. However, so far there was never any action we
needed to take other than muting that warning on or the other way.

Therefore, this commit changes it so that all pip installation calls are run
without treating warning as error.

### DIFF
--- a/testkit/_common.py
+++ b/testkit/_common.py
@@ -13,5 +13,9 @@ def run(args, env=None):
     )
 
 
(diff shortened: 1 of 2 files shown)


bigmontz (approved):
💯

PR #952, opened 2023-08-01
Fix missing lock acquisition in pool

robsdedude:
Fix some functions in the pool manipulating the routing information without holding the right lock.


bigmontz (approved):
💣

PR #955, opened 2023-08-09
Fix pool closing connections too aggressively

robsdedude:
Whenever a new routing table was fetched, the pool would close all connections to servers that were not part of the routing table. However, it might well be, that a missing server is present still in the routing table for another database. Hence, the pool now checks the routing tables for all databases before deciding which connections are no longer needed.


bigmontz (approved):
🧦

PR #961, opened 2023-09-01
Fix suppression of warnings for internal API usage

robsdedude:
Using `warnings.catch_warning` for suppressing warnings when using APIs internally that are preview/experimental/deprecated is a bad idea because the `warnings` module stores the configuration globally per module. This is not thread-safe. Therefore, the code was restructured to not rely of the `warnings` module to suppress those warnings. Instead, warning-free internal APIs are being used.

### DIFF
--- a/src/neo4j/_async/auth_management.py
+++ b/src/neo4j/_async/auth_management.py
@@ -22,7 +22,6 @@
 
 
 import typing as t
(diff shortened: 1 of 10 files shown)


gjmwoods (approved):
🧹

PR #963, opened 2023-09-11
Docs improve type hints

robsdedude:
With `from __future__ import annotations` in the code, Sphinx gets easily
confused. This applies to several areas like:
 * Type aliases
 * Things defined in `if typing.TYPE_CHECKING` blocks
 * Usage of the `typing_extensions` module

This PR aims to work around as many of those quirks as possible.

### DIFF
--- a/docs/source/async_api.rst
+++ b/docs/source/async_api.rst
@@ -235,7 +235,7 @@ Closing a driver will immediately shut down all connections in the pool.
         :param query_: cypher query to execute
         :type query_: typing.LiteralString
         :param parameters_: parameters to use in the query
-        :type parameters_: typing.Optional[typing.Dict[str, typing.Any]]
+        :type parameters_: typing.Dict[str, typing.Any] | None
         :param routing_:
             whether to route the query to a reader (follower/read replica) or
             a writer (leader) in the cluster. Default is to route to a writer.
@@ -251,7 +251,7 @@ Closing a driver will immediately shut down all connections in the pool.
                 as it will not have to resolve the default database first.
 
             See also the Session config :ref:`database-ref`.
-        :type database_: typing.Optional[str]
+        :type database_: str | None
         :param impersonated_user_:
             Name of the user to impersonate.
 
@@ -261,7 +261,7 @@ Closing a driver will immediately shut down all connections in the pool.
             permissions.
 
             See also the Session config :ref:`impersonated-user-ref`.
-        :type impersonated_user_: typing.Optional[str]
+        :type impersonated_user_: str | None
         :param auth_:
             Authentication information to use for this query.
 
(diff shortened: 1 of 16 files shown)


docs/source/async_api.rst
```diff
@@ -356,7 +354,7 @@ Closing a driver will immediately shut down all connections in the pool.
             ``parameters_``.
         :type kwargs: typing.Any
 
-        :returns: the result of the ``result_transformer``
+        :returns: the result of the ``result_transformer_``
```
bigmontz, line 357: Is a python convention?

bigmontz (approved):
✍🏼

docs/source/async_api.rst
robsdedude: That's what the parameter is called that takes the result transformer. So I decided to use the name of the parameter. Alternatively just plain text "result transformer" (without `code` formatting) would work.

src/neo4j/_sync/driver.py
```diff
@@ -715,7 +709,7 @@ def example(driver: neo4j.Driver) -> int:
         :param routing_:
             whether to route the query to a reader (follower/read replica) or
             a writer (leader) in the cluster. Default is to route to a writer.
-        :type routing_: neo4j.RoutingControl
+        :type routing_: RoutingControl
```
thelonelyvulpes, line 712: should this be?
```suggestion
        :type routing_: T_RoutingControl
```

src/neo4j/api.py
```diff
@@ -114,11 +119,8 @@ def __eq__(self, other):
 # For backwards compatibility
 AuthToken = Auth
 
-# if t.TYPE_CHECKING:
-# commented out as work around for
-# https://github.com/sphinx-doc/sphinx/pull/10880
-# make sure TAuth is resolved in the docs, else they're pretty useless
-_TAuth = t.Union[t.Tuple[t.Any, t.Any], Auth, None]
+if t.TYPE_CHECKING:
+    _TAuth = t.Union[t.Tuple[t.Any, t.Any], Auth, None]
```
thelonelyvulpes, line 123: why for the line is it using union when elsewhere you have used
`typing.Tuple[typing.Any, typing.Any] | Auth | None`

thelonelyvulpes (approved):
A couple of questions but otherwise looks good.

src/neo4j/api.py
robsdedude: Note that `|` is only being used in doc strings. That is because this typing syntax was introduced Python 3.10 ([PEP 604](https://peps.python.org/pep-0604/)). Since the driver's Python baseline is 3.7, I can't use it in code. But sphinx (the docs building tool) does still understand it if it's in docs strings. Might also be because we're building the docs with Python 3.11.

src/neo4j/_sync/driver.py
robsdedude: I think this will fail because `T_RoutingControl` is not being recognized by sphinx (the docs tool). The main point of this PR is to fix broken type info in the docs. The actual type hints remain (mostly) untouched.

PR #964, opened 2023-09-20
Fix DeprecationWarning on closing driver multiple times

robsdedude:
Closing the driver multiple times will still be allowed in future version. However, it will become a noop.


AndyHeap-NeoTech (approved):
LGTM. Worth adding a card labelled with 6.0 so we have it on the boards. Will be using those cards to remember to do deprecations etc at that time.

PR #965, opened 2023-09-20
Add support for Bolt 5.4 an API telemetry

robsdedude:
Starting with Bolt 5.4, the driver will, by default, send anonymous API usage statistics to the server if requested.

The driver configuration `telemetry_disabled=True` can be used to disable this.

```python
import neo4j

with neo4j.GraphDatabase.driver(..., telemetry_disabled=True) as driver:
    ...
```

The driver transmits the following information:

* Every time one of the following APIs is used to execute a query (for the first time), the server is informed of this (without any further information like arguments, client identifiers, etc.):
  * `driver.execute_query`
  * `session.begin_transaction`
  * `session.execute_read`, `session.execute_write`
  * `session.run`
  * the async counterparts of the above methods

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/594

### DIFF
--- a/src/neo4j/_async/io/_bolt5.py
+++ b/src/neo4j/_async/io/_bolt5.py
@@ -21,6 +21,7 @@ from enum import Enum
 from logging import getLogger
 from ssl import SSLSocket
 
+from ..._api import TelemetryAPI
 from ..._codec.hydration import v2 as hydration_v2
 from ..._exceptions import BoltProtocolError
 from ..._meta import BOLT_AGENT_DICT
@@ -167,6 +168,11 @@ class AsyncBolt5x0(AsyncBolt):
     def logoff(self, dehydration_hooks=None, hydration_hooks=None):
         """Append a LOGOFF message to the outgoing queue."""
         self.assert_re_auth_support()
+
+    def telemetry(self, api: TelemetryAPI, dehydration_hooks=None,
+                  hydration_hooks=None, **handlers) -> None:
+        # TELEMETRY not support by this protocol version, so we ignore it.
+        pass
 
     async def route(self, database=None, imp_user=None, bookmarks=None,
                     dehydration_hooks=None, hydration_hooks=None):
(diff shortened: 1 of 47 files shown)


src/neo4j/_async/io/_bolt5.py
```diff
@@ -665,3 +671,22 @@ def get_base_headers(self):
...
+
+class AsyncBolt5x4(AsyncBolt5x3):
+
+    PROTOCOL_VERSION = Version(5, 4)
+
+    def telemetry(self, api: TelemetryAPI, dehydration_hooks=None,
+                  hydration_hooks=None, **handlers) -> None:
+        if (
+            self.telemetry_disabled
+            or not self.configuration_hints.get("telemetry.enabled", False)
```
StephenCathcart, line 684: Is self.configuration_hints configured/returned from the server?

robsdedude: Yes. All configuration hints that come from the server are being stored on the connection object on `SUCCESS`ful `HELLO` https://github.com/neo4j/neo4j-python-driver/blob/fe85b713853edea0566c691aaebc6fd66c6ea9eb/src/neo4j/_async/io/_bolt5.py#L135

tests/unit/async_/io/test_class_bolt.py
```diff
@@ -65,7 +65,8 @@ def test_class_method_protocol_handlers():
         ((5, 1), 1),
         ((5, 2), 1),
         ((5, 3), 1),
-        ((5, 4), 0),
+        ((5, 4), 1),
+        ((5, 5), 0),
```
StephenCathcart, line 69: How come we need 5.5 here? Is this just future-proofing?

src/neo4j/_async/io/_bolt5.py
StephenCathcart: Thanks for clarifying!

StephenCathcart (approved):
Looks good to me! :shipit:

tests/unit/async_/io/test_class_bolt.py
robsdedude: It asserts that there are `0` Bolt 5.5 handlers registered, i.e., the driver does *not* support Bolt 5.5.

StephenCathcart: Ah ha that makes more sense, thanks.

PR #967, opened 2023-09-25
Docs: mature TCP keep alive

robsdedude:
This config option has existed for many versions. It's stable in other drivers as well. In fact, the driver didn't even issue a warning on using it. It was purely marked as experimental in the API docs. This PR aims to change this by removing the experimental mark in the docs.


AndyHeap-NeoTech (approved):
Took a while to dig through the change and understand it ;) LGTM.

PR #973, opened 2023-10-05
Propagate errors across all results in a transaction

robsdedude:
A new error `ResultFailedError` is introduced. It will be raised when using a `Result` object after the result or another result in the same transaction has failed.

User code would only ever run into this situation when catch exceptions and deciding to ignore them. Now, an error will be raised instead of undefined behavior. The undefined behavior before this fix could be (among other things) protocol violations, incomplete summary data, and hard to interpret errors.

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/598

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -1902,6 +1902,8 @@ Client-side errors
 
   * :class:`neo4j.exceptions.ResultError`
 
+    * :class:`neo4j.exceptions.ResultFailedError`
+
     * :class:`neo4j.exceptions.ResultConsumedError`
 
     * :class:`neo4j.exceptions.ResultNotSingleError`
@@ -1945,6 +1947,9 @@ Client-side errors
 .. autoexception:: neo4j.exceptions.ResultError()
     :show-inheritance:
     :members: result
+
+.. autoexception:: neo4j.exceptions.ResultFailedError()
+    :show-inheritance:
 
 .. autoexception:: neo4j.exceptions.ResultConsumedError()
     :show-inheritance:
--- a/src/neo4j/_async/work/result.py
+++ b/src/neo4j/_async/work/result.py
@@ -38,6 +38,7 @@ from ..._work import (
 )
 from ...exceptions import (
     ResultConsumedError,
+    ResultFailedError,
     ResultNotSingleError,
 )
 from ...time import (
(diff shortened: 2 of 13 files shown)


bigmontz (approved):
💯 

(However, It might be good to have some unit tests :D )

docs/source/api.rst
```diff
@@ -1946,6 +1948,9 @@ Client-side errors
     :show-inheritance:
     :members: result
 
+.. autoexception:: neo4j.exceptions.ResultFailedError()
```
thelonelyvulpes, line 1951: Understand if we have settled on a common name for this already. but is this error indicative what is actually wrong? "The result failed" is very vague, would something like: TransactionInvalidStateError be a bit more explanative? 
if the detail message could be:
`The transaction that contains this result has encountered error, all results in this transaction are therefore invalid`. if you could attach the transaction to the error so users can access it that could be useful for debugging

robsdedude: We have not defined what exact error should be thrown. So this sure is up for debate.

However, there is more detail to this massage:
 * https://github.com/neo4j/neo4j-python-driver/pull/973/files#diff-d52cb3ae0b87a7893168acf79ace03a235b5d4de3020d230e45c08fdbea54d79R470-R476
 * https://github.com/neo4j/neo4j-python-driver/pull/973/files#diff-9602a662eaafc9db0aa47c1955dafa066ef3edb2c2213439b36f9481e40ac3c3R61-R64
 
 The error name is purposefully vague because it can have multiple reasons why the result would throw it:
 * The result in question failed but the user ignored it and kept using the result
 * A different result in the same transaction failed but the user ignored it
 
 In either case, the `raise ResultFailedError(...) from ...` [here](https://github.com/neo4j/neo4j-python-driver/pull/973/files#diff-c2c44643dad938deba4854b46451154ff7147ccfdcc2f822bf567672a440b0a7R277) (specifically the `from` bit) will make the error say something like "this is a direct cause of \<insert original error with stack trace\>".

thelonelyvulpes: Thanks, should we consider exposing a property `isFailed:boolean` on the result? Whether that is in this change or subsequent one.

thelonelyvulpes (approved):

PR #975, opened 2023-10-09
Add new notification categories `SECURITY` and `TOPOLOGY`

robsdedude:


StephenCathcart (approved):
:shipit:

src/neo4j/_api.py
```diff
@@ -130,21 +130,18 @@ class NotificationDisabledCategory(str, Enum):
...
         True
         >>> NotificationDisabledCategory.DEPRECATION == "DEPRECATION"
         True
-        >>> NotificationDisabledCategory.GENERIC == "GENERIC"
-        True
 
     .. versionadded:: 5.7
 
+    .. versionchanged:: 5.14
+        Added categories :attr:`.SECURITY` and :attr:`.TOPOLOGY`.
```
thelonelyvulpes, line 143: Is it worth adding that it shouldn't be used with server versions before 5.14?

thelonelyvulpes (approved):

src/neo4j/_api.py
robsdedude: Good point. Added.

PR #989, opened 2023-11-21
Optional check for concurrent usage errors

robsdedude:
Some driver objects (e.g, Sessions, Transactions, Result streams) are not safe for concurrent use. By default, it will cause hard to interpret errors or, in the worst case, wrong behavior.

To aid finding such bugs, the driver now detects if the Python interpreter is running development mode and enables extra locking around those objects. If they are used concurrently, an error will be raised. The way this is implemented, it will only cause a one-time overhead when loading the driver's modules if the checks are disabled.

Obviously, those checks are somewhat expensive as they entail locks (less so in the async driver). Therefore, the checks are only happening if either
 * Python is started in development mode (`python -X dev ...`) or
 * The environment variable `PYTHONNEO4JDEBUG` is set (to anything non-empty) at the time the driver's modules is loaded.

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -113,6 +113,8 @@ Specifically for this driver, this will:
      It might be changed or removed any time even without prior notice.
    * the driver will raise an exception if non-concurrency-safe methods are used concurrently.
 
+     .. versionadded:: 5.15
+
 .. _development mode: https://docs.python.org/3/library/devmode.html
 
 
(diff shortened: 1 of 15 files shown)


AndyHeap-NeoTech (approved):
This is very cool stuff. I'm thinking something similar in other drivers for debug modes might be something we should do.

RichardIrons-neo4j (approved):

PR #991, opened 2023-11-22
Accept transaction config for execute_query

robsdedude:
`Driver.execute_query` now accepts a `Query` object to specify transaction config like metadata and transaction timeout. Example:

```python
from neo4j import (
    GraphDatabase,
    Query,
)

with GraphDatabase.driver(...) as driver:
    driver.execute_query(
        Query(
            "MATCH (n) RETURN n",
            # metadata to be logged with the transaction
            metadata={"foo": "bar"},
            # give the transaction 5 seconds to complete on the DBMS
            timeout=5,
        ),
        # all the other configuration options as before
        database_="neo4j",
        # ...
    )
```

Related:
 * https://github.com/neo4j-drivers/testkit/pull/607

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -187,8 +187,9 @@ Closing a driver will immediately shut down all connections in the pool.
                 query_, parameters_, routing_, database_, impersonated_user_,
                 bookmark_manager_, auth_, result_transformer_, **kwargs
             ):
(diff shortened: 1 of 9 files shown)


bigmontz (approved):
⌚

thelonelyvulpes (approved):

PR #993, opened 2023-11-27
Materialize optional Rust extension in metadata

robsdedude:
Adjust bolt agent and user agent string if Rust extensions are picked up.


AndyHeap-NeoTech (approved):
All makes sense to me.

RichardIrons-neo4j (approved):

PR #994, opened 2023-11-27
Properly mark deprecation of `get_user_agent`

robsdedude:


robsdedude:
Turns out Go and Java expose this as well. So we'll leave it.

PR #995, opened 2023-11-28
Fix generator wrapper of concurrency guard

robsdedude:
Fix up https://github.com/neo4j/neo4j-python-driver/pull/989


bigmontz (approved):
🛑

PR #1000, opened 2023-12-06
Fix Duration export in `Result.data` and `Record.data`

robsdedude:


thelonelyvulpes (approved):

tests/unit/common/test_record.py
```diff
@@ -73,17 +88,158 @@ def test_record_repr() -> None:
...
+                "👋, 🌍!",
+                [],
+                [1, 2.0, "3", True, None],
+                {"foo": ["bar", 1]},
+                b"",
+                b"foobar",
+                Date(2021, 1, 1),
+                Time(12, 34, 56, 123456789),
+                DateTime(2021, 1, 1, 12, 34, 56, 123456789),
+                Duration(1, 2, 3, 4, 5, 6, 7),
```
thelonelyvulpes, line 200: Are we missing a testcase with offsets and zone ids?

robsdedude: Could add some, yeah. It's the same class though. I'll get onto it.

PR #1004, opened 2023-12-14
Concurrency check should include session.last_bookmarks

robsdedude:


src/neo4j/_async/work/session.py
```diff
@@ -353,6 +353,7 @@ async def last_bookmark(self) -> t.Optional[str]:
             return self._bookmarks[-1]
         return None
 
+    @AsyncNonConcurrentMethodChecker.non_concurrent_method
```
bigmontz, line 356: What does this decorator means?

robsdedude: It's coming from this PR https://github.com/neo4j/neo4j-python-driver/pull/989

The driver offers an optional debug mode where it tries to detect concurrent usage of objects that are not safe to use concurrently.

bigmontz (approved):
🔢

src/neo4j/_async/work/session.py
bigmontz: thanks

PR #1006, opened 2023-12-21
Harden driver against unexpected RESET responses

robsdedude:
The server has been observed to reply with `FAILURE` and `IGNORED` to `RESET` requests. The former is according to spec and the driver should drop the connection (which it didn't), the latter isn't.

The right combination of those two unexpected responses at the right time could get the driver stuck in an infinite loop.

This change makes the driver drop the connection in either case to gracefully handle the situation.


bigmontz (approved):
🦝 

(Unit tests are more than welcome :D )

PR #1007, opened 2023-12-28
Docs: fix minimum required server version

robsdedude:
The notification category `TOPOLOGY` was introduced in Neo4j 5.13, not 5.14.

The category `SECURITY` has existed since at least 5.7, which is the minimum required version for notification filtering.

### DIFF
--- a/src/neo4j/_api.py
+++ b/src/neo4j/_api.py
@@ -151,9 +151,8 @@ class NotificationDisabledCategory(str, Enum):
     PERFORMANCE = "PERFORMANCE"
     DEPRECATION = "DEPRECATION"
     GENERIC = "GENERIC"
-    #: Requires server version 5.14 or newer.
     SECURITY = "SECURITY"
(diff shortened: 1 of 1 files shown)


src/neo4j/_api.py
```diff
@@ -151,9 +151,8 @@ class NotificationDisabledCategory(str, Enum):
     PERFORMANCE = "PERFORMANCE"
     DEPRECATION = "DEPRECATION"
     GENERIC = "GENERIC"
-    #: Requires server version 5.14 or newer.
     SECURITY = "SECURITY"
-    #: Requires server version 5.14 or newer.
+    #: Requires server version 5.13 or newer.
```
injectives, line 155: Is it worth swapping lines 155 and 156?

robsdedude: Nope. This is a doc comment, which will be picked up by the docs generation tool. It always refers to what's below.

injectives (approved):

PR #1009, opened 2024-01-08
Fix `from neo4j import *` statement

robsdedude:
The `neo4j.__all__` list got outdated breaking the star import for the top-level driver module.


StephenCathcart (approved):
LGTM 🚢

PR #1010, opened 2024-01-08
API docs: improve result stream

robsdedude:
* Show `__aiter__` and `__anext__` methods' docs string and type annotations.
 * Mention `StopAsyncIteration` exception raised by `__atier__`.
 * Same corresponding changes to the sync API.


StephenCathcart (approved):
🚢

Issue #1015, opened 2024-01-31
Pylance Typing checking error for AsyncTransaction


eric-nguyen-cs:
# Bug Report

![Screenshot 2024-01-31 at 11 52 15](https://github.com/neo4j/neo4j-python-driver/assets/77902715/96060717-a124-4359-a9fc-9906112a8bd8)

I am using the AsyncTransaction context manager

```python
 async with txn_manager as _txn:
    ...
```

However, the Pylance VSCode extension (powered by Pyright) in Basic Type checking mode is giving a type error:
`Object of type "AsyncTransaction" cannot be used with "with" because it does not implement __aenter__`
`Object of type "AsyncTransaction" cannot be used with "with" because it does not implement __aexit__`

I believe that it is because `AsyncTransaction`'s `__aenter__` and `__aexit__` are wrapped with `functools.wraps(AsyncTransactionBase._enter)` and `functools.wraps(AsyncTransactionBase._exit)` respectively

# My Environment
Python Version: 3.11
Driver Version: 5.14
Operating System: macOS

robsdedude:
Hi and thanks for reaching out.

I'm thinking this might rather be an issue with the chosen type checker (Pylance, in your case).

Reasons I'm thinking this:
 * `mypy` is happy with this
 * Looking at how `functools.wraps` is [annotated in `typeshed`](https://github.com/python/typeshed/blob/faf9d77d79adcda3bd6f8f170c7f76e30080c1d2/stdlib/functools.pyi#L115) should make the type checker infer that `wraps(xyz)` returns the identity function (only regarding `__call__`). So `wraps(xyz)(abc).__call__(...)` should be equivalent to `abc.__call__(...)`. In our example, the type checker should consider `__aenter__` and `__aexit__` to exist and be callable.
 
 Please let me know if you disagree and why.

eric-nguyen-cs:
Thank you for your quick response

I agree what you said, the bug does seem to come from `pyright` and not the `neo4j` code: I ran the code against other type checkers (`pyre-check` and `pytype`) and it seems that only `pyright` raised an issue here.

[...]

PR #1017, opened 2024-02-05
Fix internal transaction method forwarding

robsdedude:
Forwarding methods shouldn't be decorated with `functools.wraps` as it not only copies the docs string and type annotations (desirable), but also `__name__`, `__qualname__`, and more (undesirable).

While `wraps` can be configured with what properties to copy, it turns out that there is no need to keep the doc strings in the transaction base class for the forwarded methods. Instead, they are moved to the public (forwarding) methods. This still keeps the code DRY.

Suggested in: #1015

### DIFF
--- a/src/neo4j/_async/work/transaction.py
+++ b/src/neo4j/_async/work/transaction.py
@@ -18,7 +18,6 @@ from __future__ import annotations
 
 import asyncio
 import typing as t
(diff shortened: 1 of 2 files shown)


RichardIrons-neo4j (approved):
All makes sense.

PR #1018, opened 2024-02-06
Initial BenchKit backend implementation

robsdedude:


benchkit/__main__.py
```diff
@@ -0,0 +1,15 @@
...
+from sanic.worker.loader import AppLoader
+
+from .app import create_app
+from .env import env
+
+
+if __name__ == '__main__':
+    loader = AppLoader(factory=create_app)
+    app = loader.load()
+    # app.prepare(port=env.backend_port, debug=True, workers=1, dev=True)
```
bigmontz, line 15: Is it a leftover?

benchkit/app.py
```diff
@@ -0,0 +1,114 @@
...
+
+T_App: te.TypeAlias = "Sanic[Config, BenchKitContext]"
+
+
+def create_app() -> T_App:
+    app: T_App = Sanic("Python_BenchKit", ctx=BenchKitContext())
+
+    @app.main_process_start
+    async def main_process_start(app: T_App) -> None:
+        app.shared_ctx.running = Semaphore(1)
```
bigmontz, line 33: I'm curious about the function of this semaphore.

bigmontz (approved):
🦀

benchkit/__main__.py
robsdedude: No, it's for local development. The server will reload changed python files with this config. Probably worth a comment :upside_down_face: thanks.

benchkit/app.py
robsdedude: This is to make sure that only one worker process is started. The backend shouldn't use multiple drivers and each process would get its own driver instance. That's basically just a sanity check that the server is started with `workers=1` and that that option really does what it claims to do.

thelonelyvulpes (approved):

PR #1022, opened 2024-02-15
Unit tests: make work with pytest 8

robsdedude:
In pytest 8 changed the warns assertion was changed to also check for warnings if an error was raised. This used to be different, and the test suite relied (unnecessarily) on this behavior.

See also https://github.com/pytest-dev/pytest/issues/9036


RichardIrons-neo4j (approved):
🔥

PR #1034, opened 2024-04-02
[4.4] Fix handling of sub-ms transaction timeouts

robsdedude:
Transaction timeouts are specified in seconds as float. However, the server expects it in milliseconds as int. This would lead to

 1) rounding issues: previously, the driver would multiply by 1000 and
    then truncate to int. E.g., 256.4 seconds would be turned into 256399 ms
    because of float imprecision.
    Therefore, the built-in `round` is now used instead.
 2) values below 1 ms (e.g., 0.0001) would be rounded down to 0. However, 0 is
    a special value that instructs the server to not apply any timeout. This
    is likely to surprise the user which specified a non-zero timeout. In this
    special case, the driver now rounds up to 1 ms.

Back-port of: https://github.com/neo4j/neo4j-python-driver/pull/940

### DIFF
--- a/neo4j/io/_bolt3.py
+++ b/neo4j/io/_bolt3.py
@@ -226,7 +226,7 @@ class Bolt3(Bolt):
                 extra["tx_metadata"] = dict(metadata)
             except TypeError:
                 raise TypeError("Metadata must be coercible to a dict")
-        if timeout or timeout == 0:
+        if timeout or (isinstance(timeout, (float, int)) and timeout == 0):
             extra["tx_timeout"] = tx_timeout_as_ms(timeout)
         fields = (query, parameters, extra)
         log.debug("[#%04X]  C: RUN %s", self.local_port, " ".join(map(repr, fields)))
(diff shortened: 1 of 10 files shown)


neo4j/io/_bolt3.py
```diff
@@ -225,11 +226,8 @@ def run(self, query, parameters=None, mode=None, bookmarks=None,
                 extra["tx_metadata"] = dict(metadata)
             except TypeError:
                 raise TypeError("Metadata must be coercible to a dict")
-        if timeout:
-            try:
-                extra["tx_timeout"] = int(1000 * timeout)
-            except TypeError:
-                raise TypeError("Timeout must be specified as a number of seconds")
+        if timeout or timeout == 0:
```
AndyHeap-NeoTech: Not sure if I'm reading this correctly. Is timeout a float here and being used in an equality check?

AndyHeap-NeoTech (approved):
Just one question. Looks good though.

neo4j/io/_bolt3.py
robsdedude: I applied a small code change, but it can still be a comparison of a float (depending on what type the user passes to the driver). However, to my knowledge, only the only float that's falsy is `float(0)`. So the first half half of the condition would short circuit for any other value.

PR #1036, opened 2024-04-10
[4.4] Fix ITs in TeamCity

robsdedude:
For god know what reason (couldn't reproduce locally, tried every trick in the book), boltkit shutting down Neo4j after the ITs would timeout consistently in TeamCity. Since the service is (at least currently) started and stopped for the whole test session, not per test, there's little risk in just killing it instead of gracefully shutting it down.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -92,7 +92,9 @@ class Neo4jService(object):
 
     def stop(self, timeout=None):
         from shutil import rmtree
-        self.controller.stop()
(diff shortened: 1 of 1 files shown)


AndyHeap-NeoTech (approved):
A thing of beauty that we shan't mention to anyone else. Approved.

RichardIrons-neo4j (approved):

bigmontz (approved):
🔥

Issue #1024, opened 2024-02-28
Exception ignored in: <function Driver.__del__ at 0x7ff101714d30>


mfreeman451:
# Bug Report
Trying to perform a vector search in neo4j python SDK 5.17.0 and along with some results being returned, I get an exception that I can't catch.

Code: https://paste.lol/mfreeman451/neo4j-error

Error: 

```
% time python3 llm_querier.py
here is bob
yeah, bob
i would tend to agree, bob
sup bob
Exception ignored in: <function Driver.__del__ at 0x7ff101714d30>
Traceback (most recent call last):
  File "/home/mfreeman/venv/lib/python3.10/site-packages/neo4j/_sync/driver.py", line 489, in __del__
  File "/home/mfreeman/venv/lib/python3.10/site-packages/neo4j/_meta.py", line 229, in unclosed_resource_warn
TypeError: 'NoneType' object is not callable
python3 llm_querier.py  5.24s user 2.24s system 147% cpu 5.086 total
```

# My Environment
Python Version: Python 3.10.12 (main, Nov 20 2023, 15:14:05) [GCC 11.4.0]
Driver Version: Version: 5.17.0
Server Version and Edition: Neo4J 5.17 - Enterprise
Operating System: Ubuntu 22.04

mfreeman451:
fixed with `threadr_chat_vector._driver.close()` -- didn't see that in any documentations or examples..

robsdedude:
Hello and thanks for reaching out.
Great to hear you managed to figure out that you need to close the driver.

It is actually documented: https://neo4j.com/docs/api/python-driver/current/api.html#neo4j.GraphDatabase The first 2 examples of using the driver in the API docs show how to either close it using a call to `.close()`, or better yet, how to use the driver as a context manager using the `with` keyword.

mfreeman451:
I'm following along with lessons in the neo4j academy and I don't think it
was mentioned how to do it through the langchain wrapper.

adam-cowley:
Hey @mfreeman451 , which graphacademy lesson did you run into this problem on?  I will get it updated ASAP

mfreeman451:
> Hey @mfreeman451 , which graphacademy lesson did you run into this problem on? I will get it updated ASAP

https://graphacademy.neo4j.com/courses/llm-fundamentals/

PR #1037, opened 2024-04-11
Mitigate errors during resource warning on interpreter shutdown

robsdedude:
According to https://github.com/neo4j/neo4j-python-driver/issues/1024 errors can still happend during interpreter shutdown when user code is not closing the driver properly. This PR should hopefully fix this for good (1st attempt made here https://github.com/neo4j/neo4j-python-driver/pull/942).

The solution is inspired from Python's std lib. See https://github.com/python/cpython/blob/v3.12.3/Lib/multiprocessing/pool.py#L264-L271


bigmontz (approved):
🐢

PR #1045, opened 2024-04-19
[4.4] Fix: use cached port for logging

robsdedude:
Especially in the error handler, it's worth using the cached port as the underlying socket might already have been close which can lead to errors masking the actual exception that should be surfaced to the user instead.


StephenCathcart (approved):
LGTM :shipit:

thelonelyvulpes (approved):

PR #1046, opened 2024-04-23
Improve performance of `from_ordinal` of temporal types.

robsdedude:
Speeds up the function by about a factor of 100
(depending on the exact ordinal chose).


bigmontz (approved):
🍏 

(However, there is no tested related to the from_ordinal function in the code base)

robsdedude:
There actually are tests :innocent:
https://github.com/neo4j/neo4j-python-driver/blob/c0e92fb2e4f0fd8997cbd00a2c34e09f7fd1d757/tests/unit/common/time/test_date.py#L55-L89

Their exhaustiveness is very debatable though ;)

PR #1047, opened 2024-04-23
[4.4] Improve performance of `from_ordinal` of temporal types.

robsdedude:
Speeds up the function by about a factor of 100
(depending on the exact ordinal chose).

Back port of: https://github.com/neo4j/neo4j-python-driver/pull/1046


bigmontz (approved):
🦆 

(However, there is no tested related to the from_ordinal and add_months functions in the code base)

PR #1050, opened 2024-04-25
CI: fix release scripts to work with PEP 625

robsdedude:
With version 69.3.0, setuptools started naming dist archives according to PEP 625. This broke our release scripts.


bigmontz (approved):
🔐

PR #1059, opened 2024-05-08
Convenient handling of DBMS notifications

robsdedude:
This PR introduces support for a more convenient way of consuming notifications received from the DBMS server.

This feature is still in **PREVIEW**.
It might be changed without following the deprecation policy.

## Logging
A new sub-logger `neo4j.notifications` is introduced.
Every notification received will be logged through this logger.
The log-level is determined by the notification's severity.

If logging is not configured explicitly, warnings are logged to stderr.
This means that, by default, warnings like deprecations received from the DBMS will appear on stderr.

## Warnings
A new driver-level configuration `warn_notification_severity` is introduced.
It can be used to configure from which notification severity level upward the driver should emit a warning (i.e., call `warnings.warn`).

By default (`None`), it will be set to `OFF` (never emitting warnings on notifications), unless Python runs in development mode or the environment variable `PYTHONNEO4JDEBUG` is set, in which case the driver will emit a
warning on every notification.

## Usage
This functionality if mainly meant for developing and debugging.
Therefore, no emphasis is put on efficiency.
However, it's impact is capt to a minimum when disabled.
It's assumed that in productions environments this feature is either disabled explicitly

```python
neo4j.GraphDatabase.driver(
    ...,
    warn_notification_severity=neo4j.NotificationMinimumSeverity.OFF,
    ...,
)
```

or default behavior (see above) is used by not running Python in development mode.

## Example
```python
# example.py
import neo4j


URL = "neo4j://localhost:7687"
AUTH = ("neo4j", "pass")
DB = "neo4j"


with neo4j.GraphDatabase.driver(
    URL,
    auth=AUTH,
    warn_notification_severity=neo4j.NotificationMinimumSeverity.INFORMATION,
) as driver:
    driver.execute_query(
        "MERGE (:Test {name: 'foo'})",
        database_=DB,
    )
    driver.execute_query(
        "MATCH (n:Test {name: 'baz'}), (m:Test1 {name: 'bar'}) "
        "RETURN n, m",
        database_=DB,
        routing_="r",
    )
```

On an empty database, this leads to the following output:

```
$ python example.py
/path/to/example.py:10: PreviewWarning: notification warnings are a preview feature. It might be changed without following the deprecation policy. See also https://github.com/neo4j/neo4j-python-driver/wiki/preview-features.
  with neo4j.GraphDatabase.driver(
/path/to/example.py:19: Neo4jWarning: {severity: INFORMATION} {code: Neo.ClientNotification.Statement.CartesianProduct} {category: PERFORMANCE} {title: This query builds a cartesian product between disconnected patterns.} {description: If a part of a query contains multiple disconnected patterns, this will build a cartesian product between all those parts. This may produce a large amount of data and slow down query processing. While occasionally intended, it may often be possible to reformulate the query that avoids the use of this cross product, perhaps by adding a relationship between the different parts or by using OPTIONAL MATCH (identifier is: (m))} {position: line: 1, column: 1, offset: 0} for query:
MATCH (n:Test {name: 'baz'}), (m:Test1 {name: 'bar'}) RETURN n, m
^
  driver.execute_query(
Received notification from DBMS server: {severity: WARNING} {code: Neo.ClientNotification.Statement.UnknownLabelWarning} {category: UNRECOGNIZED} {title: The provided label is not in the database.} {description: One of the labels in your query is not available in the database, make sure you didn't misspell it or that the label is available when you run this statement in your application (the missing label name is: Test1)} {position: line: 1, column: 34, offset: 33} for query: "MATCH (n:Test {name: 'baz'}), (m:Test1 {name: 'bar'}) RETURN n, m"
/path/to/example.py:19: Neo4jWarning: {severity: WARNING} {code: Neo.ClientNotification.Statement.UnknownLabelWarning} {category: UNRECOGNIZED} {title: The provided label is not in the database.} {description: One of the labels in your query is not available in the database, make sure you didn't misspell it or that the label is available when you run this statement in your application (the missing label name is: Test1)} {position: line: 1, column: 34, offset: 33} for query:
MATCH (n:Test {name: 'baz'}), (m:Test1 {name: 'bar'}) RETURN n, m
                                 ^
  driver.execute_query(
```


src/neo4j/_async/work/result.py
```diff
@@ -242,17 +280,81 @@ def on_success(summary_metadata):
...
+                log_call = notification_log.info
+            elif notification.severity_level == NotificationSeverity.WARNING:
+                log_call = notification_log.warning
+            log_call(
+                "Received notification from DBMS server: %s",
+                NotificationPrinter(notification, query, one_line=True)
+            )
+
+            if notification.severity_level not in sev_filter:
+                continue
```
bigmontz, line 323: Shouldn't it be filter before logging?

src/neo4j/_async/work/result.py
```diff
@@ -242,17 +280,81 @@ def on_success(summary_metadata):
...
+                    module=globals_["__name__"],
+                    registry=warning_registry,
+                    module_globals=globals_
+                )
+
+    @property
+    def _creation_frame(self) -> t.Union[t.Literal[False], inspect.FrameInfo]:
+        if self._creation_frame_cache is not None:
+            return self._creation_frame_cache
+        if self._creation_stack is None:
```
bigmontz: Could you please put one line between the ifs? It is reading like they are the same if/else chain.

src/neo4j/_async/work/result.py
```diff
@@ -242,17 +280,81 @@ def on_success(summary_metadata):
...
+            self._creation_frame_cache = False
+        else:
+            self._creation_frame_cache = next(
+                (frame for frame in self._creation_stack
+                 if not frame.filename.startswith(str(_driver_dir))),
+                False
+            )
+        assert self._creation_frame_cache is not None  # help mypy a little
+        return self._creation_frame_cache
```
bigmontz, line 358: This will throw if self._creation_stack is None.

src/neo4j/_async/work/result.py
robsdedude: I called the config option `warn_notification_severity` hoping to convey that I only intended to be filtering what notifications cause a `warning`. It wasn't meant to affect logging. If you don't care about the logging, you can always do `logging.getLogger("neo4j.notification").disabled = True` or similar.

src/neo4j/_async/work/result.py
robsdedude: I personally don't stumble over this as if/else chains in python are if/elif chains. So you can see from the length of the keyword alone if this is continuation or not. But I can see how this might be confusing. I'll add a blank line.

src/neo4j/_async/work/result.py
robsdedude: I don't quit understand how. If `self._creation_stack` is `None`, then we end up in this branch 

https://github.com/neo4j/neo4j-python-driver/blob/2be572fe7a915255a7b046b31720b7d34b7b9e74/src/neo4j/_async/work/result.py#L347-L348

So `self._creation_frame_cache is not None` evaluates to `False` and we also return `False`.

bigmontz: Sorry, my mind tricked me.

src/neo4j/_async/work/result.py
bigmontz: So, is this PR introducing debug logs for notifications and then allowing users also to log info and warning as info and warnings?

bigmontz (approved):
📓

src/neo4j/_async/work/result.py
robsdedude: Resolved in a call.

PR #1063, opened 2024-06-11
Package driver as pure-python wheel for easier installation

### LINKED ISSUE #1061: Distribute wheel/built distribution for neo4j driver on Pypi
# Feature Request
Hi,

Would you consider releasing wheels for seqeval on Pypi as well as sdist releases?

# Pitch
This would make installing neo4j much easier for me in a specific corporate environment (at AstraZeneca, who have an enterprise license), as at the moment, the fact that neo4j requires setuptools at setup causes pip to execute without a proxy when building neo4j, as the proxy isn't passed through with PEP-517 build isolation - see this issue: https://github.com/pypa/pip/issues/7805 . If neo4j was distributed as a wheel, this would sidestep the issue (as the package would already be built when downloaded).

robsdedude:
Closes: #1061

### DIFF
--- a/bin/dist-functions
+++ b/bin/dist-functions
@@ -121,6 +121,6 @@ function setup
 {
     ARGS="$*"
     rm -rf ${DIST} 2> /dev/null
-    set_metadata_and_setup "neo4j-driver" "True" ${ARGS}   # Legacy package; can be removed in 2.0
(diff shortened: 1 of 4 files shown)


bigmontz (approved):
🐍 📦

PR #1065, opened 2024-06-14
Make usage of raw bookmarks easier

robsdedude:
(Async)Neo4jBookmarkManager's `initial_bookmarks` parameter, as well as `Bookmarks.from_raw_values` `values` parameter accept anything of type `Iterable[str]`. Unfortunately, `str` itself implements that type and yields the characters of the string. That most certainly not what the user intended.

In an ideal world, we could tell the type checker to only accept `Iterable[str]` when *not* of type `str`. See also https://github.com/python/typing/issues/256.

To help users out, we now explicitly check for `isinstance(input, str)` and turn such inputs into an iterable with the input string as the only element.


StephenCathcart (approved):
🚢

PR #1069, opened 2024-07-15
Integrate Rust extension into repository

robsdedude:
The dist scripts now generate 4 packages:

`neo4j`
Contains the driver with extensions, but falls back to pure Python
implementation if not wheel is available and compilation from source fails.
wheels + sdist are built.

`neo4j-py`
Pure Python implementation (like `neo4j` was before).
Python wheel + sdist are built.

`neo4j-rust-ext`
Enforces the existence of the Rust extensions. Else, the installation fails.
wheels + sdist are built.

`neo4j-driver`
The legacy driver alias. It's now the same as neo4j-py, except that:
only sdist is built.


robsdedude:
We've decided, for now, to just move the extensions repo to the `neo4j` GitHub org and leave the packaging as is. Python's packaging system only really knows addition. We couldn't find a clean solution to achieve our goals of:
 * `pip install neo4j` comes with extensions if possible, else falls back to pure python
 * `... neo4j-rust-ext` always comes with extensions or fails hard
 * `... some-other-package-name-tbd` enforces a pure-python driver
 
 Further, the solution should:
 * regardless of installation method not require code changes
 * leave the packages composable so that downstream libraries can add the rust extensions or pure-python enforcement simply by adding more dependencies

PR #1071, opened 2024-07-23
Fix Notification.description polyfill from GqlStatusObject

bigmontz:
Bolt 5.6 introduces the original notification description back in the protocol level. This avoids the `Notification.description` changes when connected to GQL aware servers.

This issues was detected during homologation, so the problem won't happen with any released server since the bolt version which miss information will not be released.


StephenCathcart (approved):
LGTM 👍

robsdedude (approved):
:bell::writing_hand:

PR #1087, opened 2024-08-14
API docs: Fix execute_query parameter names

robsdedude:
Trailing underscores of parameters to `Driver.execute_query` were not being properly displayed.


AndyHeap-NeoTech (approved):
All good. Approved

PR #1090, opened 2024-09-11
TestKit: fix timezone id check

robsdedude:
Correctly determine datetime values coming from TestKit that represent datetimes not existing according to the local tz db.


MaxAake (approved):
LGTM!

PR #1060, opened 2024-05-23
ADR 022: GQL compliant notifications

robsdedude:
# GQL Summary Statuses
This PR introduces GQL compliant status objects to the `ResultSummary`.

:warning: This feature is still in **PREVIEW**.
It might be changed without following the deprecation policy.

## Deprecations
`SummaryNotificationPosition` has been deprecated in favor of `SummaryInputPosition`.

## New GQL Status Objects
A new property `ResultSummary.gql_status_objects` has been introduced. It returns a sequence of `GqlStatusObject`s. These objects contain information about the execution of the query including notifications as they exist now (see `SummaryNotification`) but also an outcome status like `00000` for "success", `02000` for "no data", and `00001` for "omitted result".

Eventually, this API is planned to supersede the current notifications API.

The GqlStatusObjects will be presented in the following order:

 * A "no data" (`02xxx`) has precedence over a warning;
 * A "warning" (`01xxx`) has precedence over a success.
 * A "success" (`00xxx`) has precedence over anything informational (`03xxx`).

## Notification Filtering
Some status objects are notifications (vendor-specific definition). Those notifications can be identified using `GqlStatusObject.is_notification`. They contain additional information like a `classification`, `severity`, and `position`. Further, those notifications can be configured to be suppressed in the DBMS saving bandwidth not having to send them to the client as well as compute not having to check for those conditions.  
For this, the following driver config options can be used:

 * `notifications_min_severity` (already existing)
 * `notifications_disabled_classifications` (new, same as the existing `notifications_disabled_categories`).
   If both options are provided, they are merged into a single list of disabled classifications/categories.

All filter config options will affect both the current notifications API as well as the notifications among the GQL status objects.

## Polyfilling
There are now two APIs to access notifications: the old notifications API and the new GQL status API. To provide an easy migration path the driver will, depending on if the server is GQL aware or not (i.e., recent or old neo4j version), polyfill the API not supported by the server on a best-effort basis. In particular, GQL status objects have a GQLSTATUS code that has not counterpart in non-GQL-aware neo4j versions. Therefore, the driver will fill the GQLSTATUS with either `01N42` for warning notifications and `03N42` for informational notifications. Further, such status codes might be used by the server in the transition phase towards full GQL compliance. They indicate a lack of a final status code and will likely change in the future.

A list of all possible GQLSTATUS codes is not yet available as the server-side of this change is still very actively being worked on.

## Example Usage
```python
with neo4j.GraphDatabase.driver(
    URI,
    auth=AUTH,
[...]

PR #1093, opened 2024-09-18
Revert breaking change in summary.notification introduced in #1060

robsdedude:
When introducing bolt 5.5 support for GQL statuses in the summary, the driver gained the ability to polyfill old-style notifications from the statuses. However, `ResultSummary.notifications` used to be `None` when the server didn't send any notifications. When using the polyfill, the field would be set to `[]` instead in the same scenario. This PR fixes this.

### DIFF
--- a/src/neo4j/_work/summary.py
+++ b/src/neo4j/_work/summary.py
@@ -175,7 +175,7 @@ class ResultSummary:
 
         return notification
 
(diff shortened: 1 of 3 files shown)


MaxAake (approved):
LGTM

StephenCathcart (approved):
🚢

PR #1096, opened 2024-09-25
GQL Compliant Errors

robsdedude:
:warning: This feature is still in **PREVIEW**.
It might be changed without following the deprecation policy.

This PR introduces a new base class for `Neo4jError` and `DriverError` called `GqlError`.
It contains fields to make the errors GQL compliant.
Moreover, does the PR introduce support for Bolt protocol version 5.7 in which such GQL error information can be transmitted over the wire.

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/629


MaxAake (approved):
Looks good to me

PR #1098, opened 2024-09-25
API docs: elaborate on Result.to_df options

robsdedude:
Explain what passing `parse_dates=False` to `Result.to_df` does.


MaxAake (approved):
LGTM

PR #1099, opened 2024-09-25
Prefix AsyncNonConcurrentMethodChecker methods with underscore

robsdedude:
The concurrency check introduced in
https://github.com/neo4j/neo4j-python-driver/pull/989 introduced classmethods to several driver primitives that had the check added. These class methods are only meant for internal use and are undocumented. Therefore, they should be marked private.


MaxAake (approved):
LGTM

PR #1100, opened 2024-10-02
Fix bugs in neo4j.time.DateTime handling

robsdedude:
* Fix `DateTime` +/- `Duration` computation being wildly off by considering the days of the `DateTime` since UNIX epoch twice.
 * Fix `Result.to_df` not correctly converting `DateTime`s with `tzinfo` if `parse_dates=True` is passed in.
 * Fix `DateTime.__ne__` (inequality operator) always falling back to `object.__ne__` (comparison by `id`).


MaxAake (approved):
LGTM

PR #1101, opened 2024-10-02
[4.4] Fix bugs in neo4j.time.DateTime handling

robsdedude:
* Fix `DateTime` +/- `Duration` computation being wildly off by considering the days of the `DateTime` since UNIX epoch twice.
 * Fix `DateTime.__ne__` (inequality operator) incorrectly comparing against non-DateTime-like types.

Backport of
 * https://github.com/neo4j/neo4j-python-driver/pull/1100


MaxAake (approved):
LGTM

PR #1108, opened 2024-10-24
Improve CI

robsdedude:
* Use teamcity-messages package to make TeamCity understand test results.
* Run unit tests in parallel for each Python version to greatly reduce overall CI run-time.

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/638


StephenCathcart (approved):
👍

PR #1109, opened 2024-10-29
Improve connection error logging

robsdedude:
* Lower severity of connection error logging when closing the connection (the error is very likely to be ignored anyway)
 * Unify logging format


MaxAake (approved):
LGTM

PR #1110, opened 2024-10-30
API docs: fix `AsyncAuthManagers.basic` example

robsdedude:
The example used an old version of `ExpiringAuth`. Before version 5.9, `ExpiringAuth` accepted an `expires_in` parameter, which then was changed to `expires_at` while `expires_in` was moved to be a convenience method on `ExpiringAuth` instances.

However, the example code in `AsyncAuthManagers.basic` was not adjusted.


MaxAake (approved):
LGTM

PR #1111, opened 2024-10-31
Fix concurrent install

robsdedude:
There's some file manipulation going on during installation of the package. Since we've shifted CI to run unit tests in parallel, and therefore install the package in parallel, we've seen flakiness stemming from this race condition.


StephenCathcart (approved):
🚢

PR #1115, opened 2024-11-08
ADR 031: Homedb Cache (Bolt 5.8)

robsdedude:
Introducing support for Bolt 5.8, which (among other things) makes the server echo back the resolved home/default db on starting a transaction as well as a connection hint whether server-side routing (SSR) is enabled. If so, the echoed db can be used together with an optimistic home db cache to guess a client-side route (optimistic routing - falling back to SSR on wrong guesses). This feature saves a round-trip under the following conditions (*all* must be fulfilled):
 * Server has SSR enabled
 * Sever supports Bolt 5.8
 * A fresh (yet unused) session without an explicitly configured target database starts its first transaction (auto-commit or explicit)
 * There is a cache entry for the current user.
   How exactly "current user" is defined is left as an implementation detail. Currently, this is in order of descending precedence:
    * the impersonated user if any
    * else the session auth token (for basic auth, only the principal is considered) if any
    * else some fallback key representing the driver-level user.

Depends on (re-run CI after merge):
 * https://github.com/neo4j-drivers/testkit/pull/640
 * https://github.com/neo4j-drivers/testkit/pull/648

### DIFF
--- a/src/neo4j/_async/home_db_cache.py
+++ b/src/neo4j/_async/home_db_cache.py
@@ -49,7 +49,12 @@ class AsyncHomeDbCache:
         self._ttl = ttl
         self._cache: dict[TKey, TVal] = {}
         self._lock = AsyncCooperativeLock()
-        self._last_clean = monotonic()
+        self._oldest_entry = monotonic()
+        if max_size is not None and max_size <= 0:
+            raise ValueError(
+                f"home db cache max_size must be greater 0 or None, "
+                f"got {max_size}"
+            )
         self._max_size = max_size
 
     def compute_key(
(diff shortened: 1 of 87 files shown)


MaxAake (approved):
Looks good to me!

src/neo4j/_async/home_db_cache.py
```diff
@@ -0,0 +1,150 @@
...
+            self._oldest_entry = min(
+                (v[0] for v in self._cache.values()), default=now
+            )
+        if self._max_size and len(self._cache) > self._max_size:
+            self._cache = dict(
+                sorted(
+                    self._cache.items(),
+                    key=lambda item: item[1][0],
+                    reverse=True,
+                )[: self._truncate_size]
```
robsdedude, line 127:
```suggestion
                )[: -self._truncate_size]
```
Maybe? :eyes:

PR #1121, opened 2024-11-11
Typing: fix auth token shorthand

robsdedude:
The driver accepts a tuple `("user", "password")` for the auth argument. This is shorthand for a `basic` auth token. It was typed to expect `tuple[Any, Any]`, which is inconsistent with `neo4j.basic(str, str, ...)`.


MaxAake (approved):
LTGM 💯 💯

PR #1122, opened 2024-11-15
Set TCP_NODELAY

robsdedude:
This yields a small performance improvement by disabling Nagle's algorithm on the TCP sockets.

The change only affects the synchronous driver as most async runtime implementations already set this socket option for us.

### DIFF
--- a/src/neo4j/_async_compat/network/_bolt_socket.py
+++ b/src/neo4j/_async_compat/network/_bolt_socket.py
@@ -17,6 +17,7 @@
 from __future__ import annotations
 
 import asyncio
+import errno
 import logging
 import struct
 import typing as t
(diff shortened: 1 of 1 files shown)


src/neo4j/_async_compat/network/_bolt_socket.py
```diff
@@ -548,6 +551,12 @@ def _connect(cls, resolved_address, timeout, keep_alive):
                 s = socket(AF_INET6)
             else:
                 raise ValueError(f"Unsupported address {resolved_address!r}")
+            try:
```
AndyHeap-NeoTech, line 554: If the exception is raised does that surface to the user or get logged and the driver continues to operate using Nagels algorithm?

AndyHeap-NeoTech (approved):

AndyHeap-NeoTech (approved):
LGTM

src/neo4j/_async_compat/network/_bolt_socket.py
robsdedude: The intention is that we do this on a best-effort basis. If the socket option is not available (i.e., exception is being raised), we'll just ignore it and do what we were doing before this change.

PR #1127, opened 2024-11-28
PEP 625 compatible distributions

robsdedude:
I'll quote the mail we've received from PyPI

> This email is notifying you of an upcoming deprecation that we have determined
> may affect you as a result of your recent upload to 'neo4j-driver'.
>
> In the future, PyPI will require all newly uploaded source distribution
> filenames to comply with PEP 625. Any source distributions already uploaded
> will remain in place as-is and do not need to be updated.
>
> Specifically, your recent upload of 'neo4j-driver-5.27.0.tar.gz' is
> incompatible with PEP 625 because it does not contain the normalized project
> name 'neo4j_driver'.
>
> In most cases, this can be resolved by upgrading the version of your build
> tooling to a later version that supports PEP 625 and produces compliant
> filenames.
>
> If you have questions, you can email admin@pypi.org to communicate with the
> PyPI admin@pypi.org to communicate with the PyPI administrators.

This change allows our release job running on a recent Python version to use a newer `setuptools` version which can produce PEP 625 compatible distribution files while still allowing the package to be installed with legacy Python versions (3.7 & 3.8).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,9 @@ pyarrow = ["pyarrow >= 1.0.0"]
 
 [build-system]
 requires = [
-    "setuptools == 68.0.0",  # dropped support for Python 3.7 in 68.1.0
+    "setuptools == 68.0.0; python_version <= '3.7'",  # dropped support for Python 3.7 in 68.1.0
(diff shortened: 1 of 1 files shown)


RichardIrons-neo4j (approved):
Seems fine I guess!

AndyHeap-NeoTech (approved):

PR #1129, opened 2024-12-05
Fix coroutine not awaited warning

robsdedude:
When a certain deadline passed, the driver will not even attempt certain IO operations. By not calling the coroutine function to then ignore it under these circumstances (causing the warning), but instead deferring creation of the coroutine until it is clear we want to await it, we avoid the warning.

Functionally this changes nothing, but makes the driver less noise (and probably a better async citizen 🏅).

### DIFF
--- a/tests/unit/mixed/async_compat/test_network.py
+++ b/tests/unit/mixed/async_compat/test_network.py
@@ -37,9 +37,9 @@ if t.TYPE_CHECKING:
         TickingDateTimeFactory,
     )
 
-    TFreezeTime: te.TypeAlias = (
-        StepTickTimeFactory | TickingDateTimeFactory | FrozenDateTimeFactory
-    )
+    TFreezeTime: te.TypeAlias = t.Union[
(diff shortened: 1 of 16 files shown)


tests/unit/mixed/async_compat/test_network.py
```diff
@@ -0,0 +1,178 @@
...
+    )
+
+    TFreezeTime: te.TypeAlias = (
+        StepTickTimeFactory | TickingDateTimeFactory | FrozenDateTimeFactory
+    )
+
+
+@pytest.fixture
+def reader_factory(mocker):
+    # class MockReader(mocker.Mock):
```
MaxAake: What's the purpose of the commented out lines in this file? Is there a good chance they'll be needed later or are they just a remnant of some template?

robsdedude: The purpose is to test that a reviewer pays attention... 

<details>
<summary>Alternative explanation</summary>

Or *maaaaybe* I just forgot cleaning up properly :sweat_smile: 

</details>

MaxAake (approved):
Looks good to me!

PR #1135, opened 2025-01-13
Tests: detect timeout when joining threads

robsdedude:
It turns out that [`threading.Thread.join`](https://docs.python.org/3/library/threading.html#threading.Thread.join) does not raise an exception when the provided timeout is exceeded. This PR adjusts a unit test to accommodate that fact.

### DIFF
--- a/tests/unit/mixed/io/test_direct.py
+++ b/tests/unit/mixed/io/test_direct.py
@@ -122,7 +122,9 @@ class TestMixedConnectionPoolTestCase:
 
             # wait for all threads to release connections back to pool
             for t in threads:
-                t.join(timeout=1)
+                t.join(timeout=5)
+                if t.is_alive():
+                    raise TimeoutError(f"Joining thread timed out: {t!r}")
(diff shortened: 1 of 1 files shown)


tests/unit/mixed/io/test_direct.py
```diff
@@ -122,7 +122,9 @@ def acquire_release_conn(
 
             # wait for all threads to release connections back to pool
             for t in threads:
-                t.join(timeout=1)
+                t.join(timeout=5)
+                if t.is_alive():
```
AndyHeap-NeoTech, line 127: Showing my python ignorance here. If the timeout is triggered on the thread object then the thread remains alive?

AndyHeap-NeoTech (approved):
Looks good pending one question I had.

tests/unit/mixed/io/test_direct.py
robsdedude: That's how I read the docs, yes.

PR #1136, opened 2025-01-14
032 advertised address optimization

robsdedude:
Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/642
   (after merge, undo the temporary change of `testkit/testkit.json` and re-run CI)

### DIFF
--- a/testkit/testkit.json
+++ b/testkit/testkit.json
@@ -1,6 +1,6 @@
 {
   "testkit": {
     "uri": "https://github.com/neo4j-drivers/testkit.git",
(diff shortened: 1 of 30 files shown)


testkit/testkit.json
```diff
@@ -1,6 +1,6 @@
 {
   "testkit": {
     "uri": "https://github.com/neo4j-drivers/testkit.git",
-    "ref": "5.0"
```
robsdedude, line 4: Only temporarily! Needs to be reverted before merging the PR!

PR #1131, opened 2024-12-09
Publish wheels for legacy package, too

robsdedude:
Discussion started here: https://github.com/neo4j/neo4j-python-driver/pull/1128

PR #1139, opened 2025-02-05
Fix publishing wheels for legacy package `neo4j-driver`

robsdedude:
While #1131 enabled building the pure-Python wheels for `neo4j-driver`, it missed to update the `check-dist` and `release` scripts.


AndyHeap-NeoTech (approved):
All good.

PR #1140, opened 2025-02-06
TestKit: enable home db cache feature flag

robsdedude:
This flag was introduced later when developing the TestKit tests for the optimistic home database cache. Tests were run manually before and passed. Of course they should also run automatically in CI now.


MaxAake (approved):
👍🏻

PR #1144, opened 2025-02-18
Fix: re-auth interfering with connectivity checks

robsdedude:
Connections might have pipelined messages when being picked up from the pool (e.g., re-auth). When unflushed, those connections will be `RESET` when returned to the pool. This results in pipelineing `LOGON`, `LOGOFF`, and `RESET` where the `RESET` will skip the queue on the server side. Depending on the timing, this will leave the connection in an undesirable state (e.g., unauthenticated).

This PR circumvents this situation by acquiring an unprepared connection from the pool when performing connectivity checks. Such connections will not have any work pipelined.

### DIFF
--- a/src/neo4j/_async/work/session.py
+++ b/src/neo4j/_async/work/session.py
@@ -180,14 +180,16 @@ class AsyncSession(AsyncWorkspace):
 
     async def _get_server_info(self):
         assert not self._connection
-        await self._connect(READ_ACCESS, liveness_check_timeout=0)
+        await self._connect(
+            READ_ACCESS, liveness_check_timeout=0, unprepared=True
+        )
         server_info = self._connection.server_info
         await self._disconnect()
         return server_info
 
     async def _verify_authentication(self):
         assert not self._connection
-        await self._connect(READ_ACCESS, force_auth=True)
+        await self._connect(READ_ACCESS, force_auth=True, unprepared=True)
         await self._disconnect()
 
     @AsyncNonConcurrentMethodChecker._non_concurrent_method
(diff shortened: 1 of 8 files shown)


robsdedude:
Just a test comment.

src/neo4j/_async/work/session.py
```diff
@@ -180,14 +180,16 @@ async def _result_error(self, error):
...
+            READ_ACCESS, liveness_check_timeout=0, unprepared=True
+        )
         server_info = self._connection.server_info
         await self._disconnect()
         return server_info
 
     async def _verify_authentication(self):
         assert not self._connection
-        await self._connect(READ_ACCESS, force_auth=True)
+        await self._connect(READ_ACCESS, force_auth=True, unprepared=True)
```
MaxAake, line 192:
```suggestion
        await self._connect(READ_ACCESS, force_auth=True, unprepared=True)
```
Very ready to be wrong here, but does setting `unprepared `to True here achieve what we need? Since `force_auth `is also True it looks to me like the conditional where `unprepared `is checked would still pass?

MaxAake: Covered in a huddle, adding answer here for future reference: the `unprepared` here does not alter current behavior, but makes sure that potential future bits of pipelined work do not cause errors given that they respect `unprepared`

MaxAake (approved):
👍🏻 🫀 👨🏻‍⚕️

AndyHeap-NeoTech:
[[Python] Reset in Logon/Logoff pipelining](https://trello.com/c/MQ0Kw0SP/6013-python-reset-in-logon-logoff-pipelining)

PR #1145, opened 2025-02-20
Update link to its redirect destination

stefano-ottolenghi:


robsdedude (commented):
I'd rather not. 7687.org has been there since the beginning and I've seen the docs being shuffled around under the neo4j.com domain before. What if for instance one day, we decide to host the docs under docs.neo4j.com? I frankly think this is the more stable (and possibly also more memorable) link.

stefano-ottolenghi:
> What if for instance one day, we decide to host the docs under docs.neo4j.com?

Then there's gonna be a redirect in place from neo4j.com/docs to docs.neo4j.com, or the redirect from 7687.org would need to be updated anyway (as in "it wouldn't just keep working and it would take manual intervention"). But I don't have a strong opinion on this, so I'm gonna archive it.

PR #1149, opened 2025-03-28
[4.4] TestKit: mute setuptools warning

robsdedude:
We're not going to update the metadata of an LTS driver version. Instead we'll just mute the warning. Using a sufficiently old version of setuptools is what the user has to do in order to either get rid of the warning, or to keep the install working in the future when the deprecation has been turned into a removal.

Since we're not using a pyproject.toml on this LTS branch, we are unfortunately not in control of the version of setuptools used to install the driver.


AndyHeap-NeoTech (approved):
Approved.

PR #1152, opened 2025-04-17
TestKit: check package build is free of errors and warnings

robsdedude:


testkit/build.py
```diff
@@ -20,7 +20,11 @@
 
 
 if __name__ == "__main__":
-    run_python(["-m", "pip", "install", "-U", "pip"], warning_as_error=False)
+    run_python(
+        ["-m", "pip", "install", "-U", "pip", "build"],
```
MaxAake, line 24: Why the addition of `"build"` here? I don't think I understand what it's doing?

MaxAake (commented):
❓

testkit/build.py
robsdedude: It's installing a package called `build` which is then being invoked in the next `run_python` call to make sure that building the driver package actually works.

MaxAake: Ah, That makes sense. I'm just so used to build being a command not a package that my brain had a proper short circuit looking at it. The perils of mentally switching between ecosystems

MaxAake (approved):
✅ Looks good!

PR #1153, opened 2025-04-17
Bump minimum Python version to 3.10

robsdedude:


src/neo4j/exceptions.py
```diff
@@ -125,24 +125,15 @@
...
+    _TTransaction: t.TypeAlias = (
+        AsyncManagedTransaction
+        | AsyncTransaction
+        | ManagedTransaction
+        | Transaction
+    )
+    _TResult: t.TypeAlias = AsyncResult | Result
+    _TSession: t.TypeAlias = AsyncSession | Session
     _T = t.TypeVar("_T")
-else:
```
MaxAake, line 137: Was the code in the else block of no use? It was certainly duplicated, but this looks like a change if `t.TYPE_CHECKING` is false.

MaxAake (commented):
1️⃣ ❓

src/neo4j/exceptions.py
robsdedude: These types are in fact not being used at run-time. So there's no point in defining them.

MaxAake: I suspected as much, but not being a python wizard I wanted to get it verified.

MaxAake (approved):
✅

PR #1162, opened 2025-04-24
Raise ConfigurationError on invalid auth argument

robsdedude:
Instead of `AuthError`.

This improves the differentiation between `DriverError` for client-side errors and `Neo4jError` for server-side errors.


StephenCathcart (approved):
👍

PR #1164, opened 2025-04-25
Revamp `UnsupportedServerProduct` exception

robsdedude:
* Make it a subclass of `ServiceUnavailable`
 * Raise it (instead of internal `neo4j._exceptions.BoltHandshakeError`) when no common Bolt version could be negotiated.

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/667


MaxAake (approved):
✅Looks good to me

PR #1166, opened 2025-04-28
`ValueError` on invalid `access_mode`

robsdedude:
Raise `ValueError` if the passed `access_mode` config has an invalid value (instead of `ClientError`).


MaxAake (approved):
Looks good! ✅

PR #1168, opened 2025-04-29
`ValueError` on invalid `connection_acquisition_timeout`

robsdedude:
- `ValueError` on invalid values (instead of `ClientError`)
 - Consistently restrict the value to be strictly positive


src/neo4j/_async/io/_pool.py
```diff
@@ -1246,3 +1245,23 @@ async def on_write_failure(self, address, database):
             if table is not None:
                 table.writers.discard(address)
         log.debug("[#0000]  _: <POOL> table=%r", self.routing_tables)
+
+
+def _check_acquisition_timeout(timeout: object) -> None:
```
AndyHeap-NeoTech, line 1250: Would be neater to not have the `if timeout <= 0:` clause not repeated. I personally would not use else's for something like this. e.g.

```
if not isinstance(timeout, (float, int)):
            raise TypeError(
                "Connection acquisition timeout must be a number, "
                f"got {type(timeout)}"

if math.isnan(timeout):
            raise ValueError("Connection acquisition timeout must not be NaN")

if timeout <= 0:
            raise ValueError(
                f"Connection acquisition timeout must be > 0, got {timeout}"
            )
```

A stylistic choice mainly.

robsdedude: :+1: I was for whatever reason under the impression that `math.isnan` only takes `float`s, no `int`s. But I was wrong there. :raised_hands:

AndyHeap-NeoTech (approved):
Changes look good.

PR #1170, opened 2025-05-14
`TransactionError` on `session.run()` on busy session

robsdedude:
Raise `TransactionError` (subclass of `DriverError`) instead of `ClientError` (subclass of `Neo4jError`) when calling `session.run()` while an explicit transaction is active on that session.

 - This improves the differentiation between `DriverError` for client-side errors and `Neo4jError` for server-side errors.
 - It is now the same error raised as when trying to start an explicit transaction while another explicit transaction is already active.


AndyHeap-NeoTech (approved):
LGTM

PR #1172, opened 2025-05-14
Remove `Record.__getslice__`

robsdedude:
This magic method has been removed in Python 3.0.

If you were calling it directly, please use `Record.__getitem__(slice(...))` or simply `record[...]` instead.


AndyHeap-NeoTech (approved):
I do like to see deprecation removals

PR #1173, opened 2025-05-14
Remove `ExperimentalWarning` in favor of `PreviewWarning`

robsdedude:
- Remove `ExperimentalWarning` and turn the few left instances of it into `PreviewWarning`.
  - Deprecate importing `PreviewWarning` from `neo4j`.  
    Import it from `neo4j.warnings` instead.


tests/unit/common/warnings/test_import.py
```diff
@@ -0,0 +1,66 @@
...
+    if warning:
+        with pytest.warns(warning):
+            getattr(module, name)
+    else:
+        getattr(module, name)
+
+
+@pytest.mark.parametrize(("name", "warning"), MODULE_ATTRIBUTES)
+def test_attribute_from_import(name, warning):
+    if warning:
```
AndyHeap-NeoTech, line 52: Does the conditional change anything?

tests/unit/common/warnings/test_import.py
```diff
@@ -0,0 +1,66 @@
...
+        with pytest.warns(warning):
+            importlib.__import__("neo4j.warnings", fromlist=(name,))
```
AndyHeap-NeoTech, line 54: This seems to be the same as line 56.

AndyHeap-NeoTech (approved):

tests/unit/common/warnings/test_import.py
robsdedude: `with pytest.warns(XYZ)` asserts that the block inside emits a warning of type `XYZ`. So yeah, it's the same line twice, but once it's expected to emit a warning (this one) and once it's not (the other). Arguably this is a little moot here as `warning` is always `None` because none of the imports are deprecated or in preview. So I guess I could remove all of that logic until it's actually needed, if ever. WDYT? Leave it or remove it?

PR #1178, opened 2025-05-14
Raise `ConfigurationError` on direct driver with routing context

robsdedude:
Raise `ConfigurationError` instead of ignoring the routing context (URI query parameters) when creating a direct driver ("bolt[+s[sc]]://" scheme). A deprecation warning was already being emitted in such case.


AndyHeap-NeoTech (approved):
LGTM

AndyHeap-NeoTech (approved):

PR #1179, opened 2025-05-15
`Neo4jError` and `ClientError` deprecations and small adjustments

robsdedude:
Slightly change `Neo4jError` and `ClientError`:
  - Properties `message` and `code` are always a `str` (instead of `str | None`).
  - Remove possibility to override/set `message` and `code` properties.
  - Remove undocumented, internal methods `Neo4jError.hydrate`, `Neo4jError.invalidates_all_connections`,
    and `Neo4jError.is_fatal_during_discovery`.
  - Remove deprecated method `Neo4jError.is_retriable`.
    Use `Neo4jError.is_retryable` instead.
  - Change string representation of `Neo4jError` to include GQL error information.


tests/unit/common/test_exceptions.py
```diff
@@ -318,111 +318,260 @@ def test_error_rewrite(code, expected_cls, expected_code, mode):
...
+)
+
+
+def _expected_error_str(
+    neo4j_code: str,
+    message: str,
+    gql_status: str,
+    gql_status_description: str,
+) -> str:
+    # if not neo4j_code and mode == "gql":
```
MaxAake: This comment seems like a remnant of some WIP work. If there is some utility to keeping it around then there should be a comment explaining the meaning of it and why it is being kept.

MaxAake (commented):
💬

tests/unit/common/test_exceptions.py
MaxAake: Apologies for the poor placement of this note. It refers to the entire comment on lines 337 to 348.

robsdedude: Whoopsie :innocent: thanks for the catch :bow:

MaxAake (approved):
LGTM!

PR #962, opened 2023-09-04
Emit DeprecationWarning on using the driver after closing it

robsdedude:
Closes: #949

Issue #949, opened 2023-07-28
Missing resource closure warnings


solter:
# Bug Report
When the driver is used after it has been closed, it may open up new resources. But the `_closed` state is not reset, and so deleting a driver with open resources may leave abandoned, open resources.

## Example code
Here is an example of how this might occur:
```py
import neo4j

driver = neo4j.GraphDatabase.driver(uri, (username, password))

driver.verify_connectivity()
print(f"{driver._closed}, {len(driver._pool.connections}") # False, 1

driver.close()
print(f"{driver._closed}, {len(driver._pool.connections}") # True, 0

# del driver # this would be safe now

driver.verify_connectivity()
print(f"{driver._closed}, {len(driver._pool.connections}") # True, 1

# driver has unclosed resources
del driver # should generate an unclosed resource warning, but does not
```

# My Environment
Python Version: 3.11
Driver Version: 5.10.0
Server Version and Edition: Neo4j Desktop, Neo4j DBMS 5.9.0
Operating System: Windows 10

solter:
I'm pretty sure this is due to the following:
- the driver uses the `_close` variable to determine whether to emit the resource warning 
   - [src/neo4j/_async/driver.py Line 485](https://github.com/neo4j/neo4j-python-driver/blob/5.0/src/neo4j/_async/driver.py#L485)
- the `_close` variable is only modified by setting it to true when `close()` is called 
   - [src/neo4j/_async/driver.py Line 554](https://github.com/neo4j/neo4j-python-driver/blob/5.0/src/neo4j/_async/driver.py#L554)

Thus if resources are allocated (for example in the `verify_connectivity` method) _after_ the driver has been closed, the `_closed` state is not reset.

Potential solutions:
1. update the resource warning check to fire if any resources are allocated in the `_pool` object
   - should just be able to check the `_pool` since it is the only thing cleaned up by `close()`
   - instead of when the `_closed` flag is set
   - this is probably the most robust fix
      - assuming that it is known how to query to pool for open resources for any kind of pool
         - unclear if this is possible, since the type declaration states that `_pool` is of type any
2. update the driver to disallow method calls that might open resources after it has been closed
   - could break existing code
   - would not recommend
3. update the `_closed` flag so that it is reset to false any time a method is called that may allocate resources
   - probably the easiest fix
   - should be similarly robust as method 2, but slightly more error prone for future development
      - since any new methods on Driver or Driver subclasses would need to remember to reset the `_close` flag

[...]

PR #1182, opened 2025-05-16
Fail hard on using closed driver

robsdedude:
* Raise `DriverError` on using the closed driver.
 * Calling `driver.close()` again is now a no-op.
 
 Related: #949, #962


StephenCathcart (approved):
Looks good

PR #1183, opened 2025-05-16
Don't close drivers & sessions in finalizer

robsdedude:
No longer implicitly closing drivers and sessions in `__del__()` (finalizer/destructor). Make sure to call `.close()` on them explicitly or use them in a `with` statement.

A deprecation warning has been emitted in such cases before.

Relying on `__del__` leads to hard to follow control flow and can result in surprising behavior as there is no guarantee when that the finalizer will be called if ever. A ResourceWarning is emitted instead.

See also: [https://docs.python.org/3/reference/datamodel.html#object.\_\_del\_\_](https://docs.python.org/3/reference/datamodel.html#object.__del__)


StephenCathcart (approved):
👍

PR #1184, opened 2025-05-16
Remove `session.read/write_transaction`

robsdedude:
Remove deprecated `session.read_transaction` and `session.write_transaction`.
Instead, use `session.execute_read` and  `session.execute_write` respectively.


MaxAake (approved):
Looks good!

PR #1185, opened 2025-05-16
Make `Summary.summary_notifications` a `tuple`

robsdedude:
Make `Summary.summary_notifications` a `tuple` instead of a `list` and type it with `Sequence` to signify that it should be treated as immutable.


MaxAake (approved):
Looks good! ✅

PR #1187, opened 2025-05-16
Drop EntitySetView indexing by `int` id

robsdedude:
Graph type sets (`neo4j.graph.EntitySetView`) can no longer by indexed by legacy `id` (`int`, e.g., `graph.nodes[0]`).  
Use the `element_id` instead (`str`, e.g., `graph.nodes["..."]`).


AndyHeap-NeoTech (approved):
LGTM

PR #1192, opened 2025-05-20
Fix driver stuck on RecursionError on COMMIT SUCCESS

robsdedude:
The driver could get stuck when the result summary was so deeply nested that it caused an `RecursionError`. It would also get stuck on certain protocol violations in the same place (e.g., unknown PackStream tag). However, the latter is much more unlikely to occur.

This patch will make the driver ditch the connection when it's not able to properly read a message for unexpected reasons and let the error bubble up.


tests/unit/async_/io/_common/test_inbox.py
```diff
@@ -0,0 +1,140 @@
...
+            bytes((0, 2, 10, 11, 0, 2, 12, 13, 0, 1, 14, 0, 0)),
+            bytes(range(10, 15)),
+        ),
+        (
+            bytes((0, 2, 10, 11, 0, 2, 12, 13, 0, 0)),
+            bytes(range(10, 14)),
+        ),
+        (
+            bytes((0, 1, 5, 0, 0)),
+            bytes((5,)),
```
MaxAake, line 77:
```suggestion
            bytes((5)),
```
Nit: trailing comma on argument. also in test_outbox, won't comment again.

MaxAake (approved):
✅ 💬

tests/unit/async_/io/_common/test_inbox.py
robsdedude: :upside_down_face: Tis an amazing Python footgun. :gun: 

```python
>>> a = (5); print(a, type(a))
5 <class 'int'>
>>> a = (5,); print(a, type(a))
(5,) <class 'tuple'>
```

PR #1193, opened 2025-05-20
Remove deprecated `ServerInfo.connection_id`.

robsdedude:
There is no replacement as this is considered internal information.


StephenCathcart (approved):
🚢

PR #1175, opened 2025-05-14
Remove `Bookmark` in favor or `Bookmarks`

robsdedude:
- Remove deprecated class `neo4j.Bookmark` in favor of `neo4j.Bookmarks`.
- Remove deprecated class `session.last_bookmark()` in favor of `last_bookmarks()`.

PR #1196, opened 2025-05-22
Deprecate raw `initial_bookmarks` for `bookmark_manager`

robsdedude:
Deprecate passing raw sting bookmarks as `initial_bookmarks` to `GraphDatabase.bookmark_manager()`.
Use a `neo4j.Bookmarks` object instead.

Amends https://github.com/neo4j/neo4j-python-driver/pull/1175


AndyHeap-NeoTech (approved):
Put it in a box and ship it.

PR #1197, opened 2025-05-22
Refactor type annotations

robsdedude:
* Move all imports for typing into a central file.  
   This will make upgrading easier when bumping the minimum Python version in the future. It also enables a more consistent code style.
 * Upgrade typing imports to be in line with the new Python 3.10 base-line.  
   E.g., import `collections.abc.Mapping` instead of `typing.Mapping`.
 * Moved some imports out of `if TYPE_CHECKING` blocks to help Sphinx produce useful type annotations in the docs again.


MaxAake (approved):
✍🏻 ✅

PR #1198, opened 2025-05-27
TestKit backend: remove `severity` fill from summary serialization

robsdedude:
The driver API never had this field, so the backend was faking it.

Said API has been deprecated in other drivers and was now removed in 6.x. Therefore, the backend fill can also be removed. See also:

 * https://github.com/neo4j-drivers/testkit/pull/673


RichardIrons-neo4j (approved):
outstanding work

PR #1199, opened 2025-05-27
Fix: make unsupported comparisons return `NotImplemented`

robsdedude:
Make all comparator magic methods return `NotImplemented` instead of `False` (or raising `TypeError` in some instances) if the other operand is not of a supported type.

This means that when comparing a driver type with another type is doesn't support, the other type get the chance to handle the comparison.

Affected types:
 * `neo4j.Record`
 * `neo4j.graph.Node`, `neo4j.graph.Relationship`, `neo4j.graph.Path`
 * `neo4j.time.Date`, `neo4j.time.Time`, `neo4j.time.DateTime`
 * `neo4j.spatial.Point` (and subclasses)


AndyHeap-NeoTech (approved):
Looks correct.

PR #1202, opened 2025-05-28
[5.28] Fix driver stuck on RecursionError on COMMIT SUCCESS

robsdedude:
The driver could get stuck when the result summary was so deeply nested that it caused an `RecursionError`. It would also get stuck on certain protocol violations in the same place (e.g., unknown PackStream tag). However, the latter is much more unlikely to occur.

This patch will make the driver ditch the connection when it's not able to properly read a message for unexpected reasons and let the error bubble up.

Backport of: https://github.com/neo4j/neo4j-python-driver/pull/1192

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -281,6 +281,9 @@ select = [
     # allow async functions without await to enable type checking, pretending to be async, matching type signatures
     "RUF029",
 ]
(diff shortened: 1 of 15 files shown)


MaxAake (approved):
🍒 👨🏻

PR #1204, opened 2025-06-02
Bump pre-commit hooks & fix new lints

robsdedude:


tests/unit/async_/work/test_result.py
```diff
@@ -1388,8 +1388,8 @@ async def test_notification_warning(
             await result._run("CYPHER", {}, None, None, "r", None, None, None)
             await result.consume()
     else:
+        await result._run("CYPHER", {}, None, None, "r", None, None, None)
```
MaxAake: as per in-person discussion: might be better to ignore this lint

MaxAake (approved):
✅

PR #1206, opened 2025-06-03
Introduce sub-logger `neo4j.session`

robsdedude:
Separate out log entries that are session-related (including transaction retries) form sub-logger `neo4j.pool` to a new sub-logger `neo4j.session`.


StephenCathcart (approved):
👍

PR #1207, opened 2025-06-05
PEP 639 compliant licensing meta data

robsdedude:
https://peps.python.org/pep-0639/#deprecate-license-classifiers


AndyHeap-NeoTech (approved):
LGTM

PR #1211, opened 2025-06-12
Make DNS error retryable

robsdedude:
Configuring the driver with a URL that cannot be DNS resolved will raise a (retryable) `ServiceUnavailable` error instead of a `ValueError`.


src/neo4j/_addressing.py
```diff
@@ -200,7 +200,7 @@ def parse_list(
 
             >>> Address.parse_list("localhost:7687", "[::1]:7687")
             [IPv4Address(('localhost', 7687)), IPv6Address(('::1', 7687, 0, 0))]
-            >>> Address.parse_list("localhost:7687 [::1]:7687")
+            >>> Address.parse_list("localhost:7687", "[::1]:7687")
```
MaxAake: This looks like it makes lines 203-204 redundant (identical to 201-202). and the comment on 198-199 implies that the earlier version of the code was a valid usage example.

MaxAake (commented):
💬

MaxAake (approved):
✅

PR #1213, opened 2025-06-24
TestKit: skip connection acquisition timeout unification tests for now

robsdedude:
Skip tests adjusted in https://github.com/neo4j-drivers/testkit/pull/674 until we get around to implement the pending unification.


MaxAake (approved):
👍🏻

PR #1214, opened 2025-06-25
Bolt 6.0 support with vector types

robsdedude:
Introduce Bolt 6.0 support, which includes vector types.

Example:
```python
from neo4j.vector import Vector

# Create a vector (raw big-endian byte buffer by default)
v = Vector(b"\x00\x01\x02\x03", "i16")
# or (less performant, but much more user-friendly)
v = Vector.from_native([1, 513], "i16")
# `v` can now be used as query parameter

# Get a vector
v = driver.execute_query(
    "MATCH (d:Doc) RETURN d.embedding AS v LIMIT 1",
    database_="neo4j",
).records[0]["v"]
# Working with the vector
print(f"Got vector of type {v.dtype} with {len(v)} elements: {v}")
print("raw bytes:", v.raw())
print("python list:", v.to_native())
print("as numpy array:", v.to_numpy())
```


docs/source/index.rst
```diff
@@ -9,8 +9,8 @@ Bolt protocol versions supported:
 
 .. # [bolt-version-bump] search tag when changing bolt version support
 
+* Bolt 6.0
 * Bolt 5.0 - 5.8
-* Bolt 4.4
```
MaxAake, line 13: This should be put on hold pending the discussion on documenting support for 4.4

tests/unit/common/vector/test_vector.py
```diff
@@ -0,0 +1,856 @@
...
+    _mock_mask_extensions(mocker, ext)
+
+    print(timeit.timeit(data.to_native, number=1_000))  # noqa: T201
+    print(timeit.timeit(data.to_native, number=1_000))  # noqa: T201
+
+    print()  # noqa: T201
+    data = Vector.from_bytes(bytes(i % 256 for i in range(8 * 1)), dtype)
+    print(timeit.timeit(data.to_native, number=100_000))  # noqa: T201
+    print(timeit.timeit(data.to_native, number=100_000))  # noqa: T201
+
```
MaxAake: is there a reason that the timeit command is run twice for each data here?

MaxAake (commented):
Two minor comments

tests/unit/common/vector/test_vector.py
robsdedude: Nope, copy pasta mistake. Thanks.

MaxAake (approved):
✅

tests/unit/common/vector/test_vector.py
robsdedude: I decided to completely remove the benchmarks and have them be part of the rust-ext repo where they make more sense anyway.

See https://github.com/neo4j/neo4j-python-driver-rust-ext/pull/45/commits/e99e5352e4fad3ca85e84a080cfdbeb0733b94b3

PR #1216, opened 2025-07-14
Introduce `GqlError.find_by_gql_status`

robsdedude:
Return the first GqlError in the cause chain with the given GQL status.

This method traverses this GQLErorrs's :attr:`__cause__` chain, starting with this error itself, and returns the first error that has the given GQL status. If no error matches, :data:`None` is returned.

Example:

```python
def invalid_syntax(err: GqlError) -> bool:
    return err.find_by_gql_status("42001") is not None
```

### DIFF
--- a/src/neo4j/exceptions.py
+++ b/src/neo4j/exceptions.py
@@ -219,6 +219,14 @@ class GqlError(Exception):
     Instead, only subclasses are raised.
     Further, it is used as the :attr:`__cause__` of GqlError subclasses.
 
+    Sometimes it is helpful or necessary to traverse the cause chain of
+    GQLErrors to fully understand or appropriately handle the error. In such
+    cases, users can either traverse the :attr:`__cause__` attribute of the
+    error(s) or use the helper method :meth:`.find_by_gql_status`. Note that
+    :attr:`__cause__` is a standard attribute of all Python
+    :class:`Exception`s. Therefore, the cause chain may also contain other
+    types besides GqlError.
+
     .. versionadded: 5.26
 
     .. versionchanged:: 6.0 Stabilized from preview.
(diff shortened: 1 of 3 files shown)


src/neo4j/exceptions.py
```diff
@@ -219,6 +219,14 @@ class GqlError(Exception):
     Instead, only subclasses are raised.
     Further, it is used as the :attr:`__cause__` of GqlError subclasses.
 
+    Sometimes it is helpful or necessary to traverse the cause chain of
+    GQLErrors to fully understand or appropriately handle the error. In such
+    cases, users can either traverse the :attr:`__cause__` attribute of the
+    error(s) or use the helper method :meth:`.find_by_gql_status`. Note that
+    :attr:`__cause__` is a standard attribute of all Python
+    :class:`Exception`s. Therefore, the cause chain may also contain other
```
stefano-ottolenghi:
```suggestion
    :class:`Exception`s: the cause chain may therefore contain other
```

tests/unit/common/test_exceptions.py
```diff
@@ -868,3 +868,33 @@ def test_deprecated_setter(attr):
...
+def test_find_by_gql_status(insert_after: int) -> None:
+    error_to_find = _make_test_gql_error("12345")
+
+    root = None
+    if insert_after == -1:
+        root = error_to_find = _make_test_gql_error("12345")
+    for i in range(3):
+        root = _make_test_gql_error(f"{i + 2}2345", cause=root)
+        if i == insert_after:
+            root = error_to_find = _make_test_gql_error("12345", cause=root)
```
MaxAake: This test functions but is a bit hard to follow, as established with an in-person talk, it would probably be easier to handle `error_to_find` much like `root` is handled.

MaxAake (approved):
🔎 ✅

PR #1218, opened 2025-07-15
[5.28] Docs: specify applicability of `is_retryable`

robsdedude:
Because the server-side state of auto-commit transactions cannot be known by the driver (they might even be partially committed), it's not safe to retry them solely based on the exception observed. Business logic is required to determine if the auto-commit transaction succeeded or failed after observing an error client-side.

Backport of https://github.com/neo4j/neo4j-python-driver/pull/1217


MaxAake (approved):
✅

PR #1224, opened 2025-07-21
Make `UnsupportedServerProduct` subclass of `ConfigurationError`

robsdedude:
Amends https://github.com/neo4j/neo4j-python-driver/pull/1164

Reconsidered the change.
 * The error shouldn't be retryable. If driver and server cannot agree on a common bolt version, it's highly unlikely they will a short moment later.
 * `ServiceUnavailable` is used for connectivity issues (broken sockets, failed dial up, etc.). This error is not really in that category.


RichardIrons-neo4j (approved):
All good!

PR #1225, opened 2025-07-23
API Docs: remove Bolt 4.4 support

robsdedude:
6.x drivers may or may not implement Bolt 4.4. If they do, it might get removed anytime if it becomes a maintenance burden. It's an undocumented and unsupported feature.


StephenCathcart (approved):
🎉

PR #1226, opened 2025-07-24
Vector types: optimize error path

robsdedude:
The pure-python implementations of `Vector.from_native` perform type checks. This patch makes the driver fail fast when passing incompatible types instead of scanning the whole Iterable parameter.


StephenCathcart (approved):
👍

PR #1227, opened 2025-07-25
Fix socket timeouts

robsdedude:
## Fix connection hint timeout
The driver may receive a configuration hint from the server during connection setup. The hint is a timeout: if the driver is waiting for a response from the server but has not heard back within the timeout window, the driver shall assume the connection is dead. The server can send `NOOP` ping messages to reset the timer ticking on the client-side.

The driver applied this hint to both reads and writes. This PR changes it so that the hint, as intended, only applies to read operations on connections.

## New write timeout configuraiton option
To avoid the driver hanging indefinitely on write operations, a new driver-level configuration option `connection_write_timeout`, which defaults to `30 seconds` is introduced. As the name suggests, this timeout is applied to all write operations on connections.


StephenCathcart (approved):
👍 looks good to me

PR #1231, opened 2025-08-21
Improve repr string representation of spatial types

robsdedude:
Instead of `POINT(1.0 2.0 3.0)` (regardless of the sub-type) the `repr` string representation is now `WGS84Point(1.0, 2.0, 3.0)` or similar.

This is more in line with Python's recommendations:

> 


MaxAake (approved):
✅

PR #1232, opened 2025-08-21
Improve packstream Structure class

robsdedude:
These changes have no affect on the driver's public API. They're targeted at improving the development and debugging experience.

 * Adjust `repr` to follow Python's recommendations
 * Fix `__eq__` returning `NotImplementedError` instead of `NotImplemented`
 * Add type hints
 * Add tests


MaxAake (approved):
✅ 
❓ What is the rules for including changes in the changelog in the python driver? I'm not sure if the lack of a changelog update here is an oversight or deliberate

robsdedude:
> What is the rules for including changes in the changelog in the python driver?

Great question. Firstly, there are 2 change logs:
 * https://github.com/neo4j/neo4j-python-driver/blob/6.x/CHANGELOG.md
 * and https://github.com/neo4j/neo4j-python-driver/wiki#changelog
 
The latter is a full changelog with anything that's user-facing. The former is only major and breaking changes. Whatever that means. When I took over maintenance of this project, this was already like that. It might be worth tackling that at some point.

Either way, the user should under normal conditions never end up with a `Structure` object in their hand. Therefore, this PR doesn't alter the public API or the driver's behavior except for maybe in log messages or buggy situations. So I think it's not worth being included in either changelog.

PR #1233, opened 2025-08-21
Improve repr string representation of `SummaryCounters`

robsdedude:
Instead of for example `{'nodes_created': 42, 'labels_removed': 0, '_contains_system_updates': True}` the `repr` string representation is now `SummaryCounters({'nodes-created': 42, 'labels-removed': 0, 'contains-system-updates': True})` or similar.

This is more in line with Python's recommendations:

> 

Further, the `str` representation has been adjusted from for instance `{'nodes_created': 42, 'labels_removed': 0, '_contains_system_updates': True}` to  `SummaryCounters{nodes_created: 42, contains_updates: True, contains_system_updates: True}` representing all non-default attributes/properties of the object.

### DIFF
--- a/tests/unit/common/work/test_summary.py
+++ b/tests/unit/common/work/test_summary.py
@@ -867,7 +867,7 @@ def test_summary_result_counters(summary_args_kwargs, counters_set) -> None:
             "SummaryCounters({})",
         ),
         (
-            {"unreckognized": 42},
+            {"unrecognized": 42},
             "SummaryCounters({})",
         ),
         (
@@ -1051,7 +1051,7 @@ def test_summary_counter_repr(stats: dict, expected: str) -> None:
             ),
         ),
         (
-            {"unreckognized": 42},
+            {"unrecognized": 42},
             (
                 "SummaryCounters{"
                 "contains_updates: False, "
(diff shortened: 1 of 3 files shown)


tests/unit/common/work/test_summary.py
```diff
@@ -859,6 +859,533 @@ def test_summary_result_counters(summary_args_kwargs, counters_set) -> None:
...
 
+@pytest.mark.parametrize(
+    ("stats", "expected"),
+    (
+        (
+            {},
+            "SummaryCounters({})",
+        ),
+        (
+            {"unreckognized": 42},
```
MaxAake:
```suggestion
            {"unrecognized": 42},
```

tests/unit/common/work/test_summary.py
```diff
@@ -859,6 +859,533 @@ def test_summary_result_counters(summary_args_kwargs, counters_set) -> None:
...
+            {},
+            (
+                "SummaryCounters{"
+                "contains_updates: False, "
+                "contains_system_updates: False"
+                "}"
+            ),
+        ),
+        (
+            {"unreckognized": 42},
```
MaxAake:
```suggestion
            {"unrecognized": 42},
```

MaxAake (approved):
✅ ✏️ 
Just a minor spelling error in tests

PR #1234, opened 2025-08-22
Improve repr string representation of `GqlStatusObject`

robsdedude:
Instead of for example 
```
GqlStatusObject(gql_status='12345', status_description='descr', position=<SummaryInputPosition object at ...>, raw_classification='HINT', classification=<NotificationClassification.HINT: 'HINT'>, raw_severity='WARNING', severity=<NotificationSeverity.WARNING: 'WARNING'>, diagnostic_record={...})
```
the `repr` string representation is now 
```
<GqlStatusObject gql_status='12345', status_description='descr', position=<SummaryInputPosition line=1, column=2, offset=3>, raw_classification='HINT', classification=<NotificationClassification.HINT: 'HINT'>, raw_severity='WARNING', severity=<NotificationSeverity.WARNING: 'WARNING'>, diagnostic_record={...}>
```
or similar.

This is more in line with Python's recommendations since `GqlStatusObject` has no such public constructor:

> 

To enable the above change, `SummaryInputPosition`'s `repr` string representation was also changed to contain more information.


MaxAake (approved):
✅

PR #1235, opened 2025-08-22
Improve repr string representation of `Duration` and `ClockTime`

robsdedude:
Instead of for example `Duration(months=1, days=2, seconds=3, nanoseconds=4)` the `repr` string representation is now `neo4j.time.Duration(months=1, days=2, seconds=3, nanoseconds=4)` or similar.

Both are fine in regards to Python's recommendations. However, the latter is more consistent with all other types defined in `neo4j.time`.

`ClockTime`'s `repr` was changed in the same way.


MaxAake (approved):
✅

PR #1236, opened 2025-08-22
Improve repr string for `OrderedSet` and `RoutingTable`

robsdedude:
Both types are not public. Therefore, this change won't affect public driver API. It might, however, show up in driver logs.


MaxAake (approved):
✅

PR #1239, opened 2025-09-12
API docs: explain security config options interplay

### LINKED ISSUE #1237: Better explanation of using custom/self-signed certificates
# Feature Request

I'm trying to understand how to properly use Custom Certificates (or Self Signed Certificates).
The documentation and related error messages are confusing.

The documentation here says that if I want to use a custom/self-signed certificate I need to use the `neo4j+ssc` scheme.
https://neo4j.com/docs/api/python-driver/5.28/api.html#uri

However when I do that I get an error.

`neo4j.exceptions.ConfigurationError: The config settings "encrypted", "trust", "trusted_certificates", and "ssl_context" can only be used with the URI schemes ['bolt', 'neo4j']. Use the other URI schemes ['bolt+ssc', 'bolt+s', 'neo4j+ssc', 'neo4j+s'] for setting encryption settings.`

I've tried using both SSL `ssl_context` and `trusted_certificates`, they both throw the same error.

MRE(ish) below 👇 

```python
import pathlib
import ssl

import neo4j

from my_app import CUSTOM_CERT_PATH # pathlib.Path object

def get_config() -> dict:
   return {"user_name": "username", "password": "pw", "uri": "neo4j+ssc://12345.databases.neo4j.io"}

def get_neo4j_driver_ssl_context(config: dict[str, str]) -> neo4j.Driver:
    return GraphDatabase.driver(
        config["uri"],
        auth=(config["username"], config["password"]),
        ssl_context=ssl.create_default_context(cafile=CUSTOM_CERT_PATH),
    )

def get_neo4j_driver_custom_ca(config: dict[str, str]) -> neo4j.Driver:
    return GraphDatabase.driver(
        config["uri"],
        auth=(config["username"], config["password"]),
        trusted_certificates=neo4j.TrustCustomCAs(str(CUSTOM_CERT_PATH))
    )

if __name__ == "__main__":
  config = get_config()
  ssl_ctx_driver = get_neo4j_driver_ssl_context(config)
  ssl_ctx_driver.verify_connectivity()

  custom_ca_driver = get_neo4j_driver_custom_ca(config)
  custom_ca_driver.verify_connectivity()
```

I should note I've used this exact cert file with multiple other (non Aura/Neo4J) services without issue.

The DB in question is on/in Aura if the uri didn't make that clear.

I can of course connect when I drop the the `+ssc` scheme (example: `neo4j+ssc://12345.databases.neo4j.io`) but it goes without saying I don't want to to use the driver without encryption.

### Version Info
```
neo4j v5.28.2
Python 3.13.5
```

# Pitch

Please update or expand your documentation around using custom certificates along with examples, or fix the underlying issue that prevents using self signed certs in the manor that is described.

robsdedude:
Fixes https://github.com/neo4j/neo4j-python-driver/issues/1237

Closes DRIVERS-30

### DIFF
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -97,7 +97,9 @@ Each supported scheme maps to a particular :class:`neo4j.Driver` subclass that i
 
 
 .. note::
+
     See also :ref:`encryption-config-note-ref` to understand how the URI scheme relates to other encryption configuration options.
+
 
 .. note::
 
@@ -809,7 +811,7 @@ There are different *mutually exclusive* ways of configuring TLS/SSL encryption 
 
 * Use a URI scheme ending in ``+s``. This auto-configures the driver to use TLS and only trust system CAs.
 * Use a URI scheme ending in ``+ssc``. This auto-configures the driver to use TLS and trust any certificate.
-* Use any other URI scheme and
+* Use a URI scheme without suffix (i.e. ``neo4j://`` or ``bolt://``) and one of the following mutually exclusive options:
 
   * set :ref:`encrypted-ref` to ``True`` and optionally configure :ref:`trusted-certificates-ref` and/or
     :ref:`client-certificate-ref` to enable TLS with custom security settings.
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -603,6 +603,9 @@ For example:
 ---------------
 Specify a custom SSL context to use for wrapping connections.
 
+The driver offers other, easier APIs for common encryption configurations (see :ref:`encryption-config-note-ref`).
+It's likely that your use-case doesn't actually require this options.
+
 This setting is only available for URI schemes ``bolt://`` and ``neo4j://`` (:ref:`uri-ref`).
 
 If given, ``encrypted``, ``trusted_certificates``, and ``client_certificate`` have no effect.
--- a/docs/source/api.rst
+++ b/docs/source/api.rst
@@ -95,6 +95,10 @@ Each supported scheme maps to a particular :class:`neo4j.Driver` subclass that i
 | neo4j+s                | :ref:`neo4j-driver-ref` with encryption (accepts only certificates signed by a certificate authority), full certificate checks.       |
 +------------------------+---------------------------------------------------------------------------------------------------------------------------------------+
 
+
+.. note::
+    See also :ref:`encryption-config-note-ref` to understand how the URI scheme relates to other encryption configuration options.
+
 .. note::
 
     See https://neo4j.com/docs/operations-manual/current/configuration/ports/ for Neo4j ports.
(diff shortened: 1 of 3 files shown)


docs/source/api.rst
```diff
@@ -794,6 +801,23 @@ The driver transmits the following information:
...
 
+.. _encryption-config-note-ref:
+
+Note on Encryption Configuration
+--------------------------------
+There are different *mutually exclusive* ways of configuring TLS/SSL encryption behavior of the driver:
+
+* Use a URI scheme ending in ``+s``. This auto-configures the driver to use TLS and only trust system CAs.
+* Use a URI scheme ending in ``+ssc``. This auto-configures the driver to use TLS and trust any certificate.
+* Use any other URI scheme and
```
stefano-ottolenghi: I believe the `any` bit refers to `neo4j://` vs `bolt://`, but I think it's worth making it explicit with something like

> A URI scheme without suffix (i.e. `neo4j://` or `bolt://`)

Also I'd clarify that the following options are mutually exclusive, hence my final suggestion

> Use a URI scheme without suffix (i.e. `neo4j://` or `bolt://`) and one of the following mutually exclusive options:

docs/source/api.rst
```diff
@@ -794,6 +801,23 @@ The driver transmits the following information:
...
+
+  * set :ref:`encrypted-ref` to ``True`` and optionally configure :ref:`trusted-certificates-ref` and/or
+    :ref:`client-certificate-ref` to enable TLS with custom security settings.
+  * or set :ref:`ssl-context-ref` to gain full control (and responsibility) over the TLS configuration.
+  * or set ``encrypted=False`` (default) to disable TLS.
```
stefano-ottolenghi, line 822: How does option 1 differ from 2? What can I do with 2 that I can't do with 1? Can we given a hint in a few words?

robsdedude: This is a bottomless pit. You can do pretty much anything with 2. From simple things like limiting which TLS versions or encryption ciphers to allow all the way to mocking the SSL context for dirty hacks like intercepting/monitoring all traffic. Also, I think it's desirable to not be to comprehensive in explaining the ssl_context option. For several reasons:
 * It's pretty advanced, and using it it's easy to end up with an insecure config (there's a warning somewhere else). Probably only users that understand Python's ssl lib well should use it.
 * It's pretty niche. The driver offers easier APIs for common TLS configurations. Using a custom ssl context should be the last resort.

If you have a look at https://docs.python.org/3/library/ssl.html#ssl-contexts you'll see all the things you can configure on a standard SSLContext. And then, as mentioned, nothing stops the user from implementing their own duck-typed SSLContext at which point you can do pretty much whatever you want.

AndyHeap-NeoTech: It might be worth mentioning the advanced nature and that a good understanding of the ssl lib is recommended.

AndyHeap-NeoTech (approved):
Looks good, just one comment.

docs/source/api.rst
robsdedude: There's also a link to `ssl-context-ref` which contains said warning. I suggest to add any additional info about this config option there. Because that's where I expect people to look if they want to learn more about it.

stefano-ottolenghi (approved):

AndyHeap-NeoTech (approved):

PR #1209, opened 2025-06-10
Deprecate notifications, stabilize GQL statuses in turn

robsdedude:
* Notifications
   * Deprecate notifications and related APIs:
     * `ResultSummary.notifications`
     * `ResultSummary.summary_notifications`
     * `neo4j.SummaryNotification`
     * `neo4j.NotificationCategory`
     * `neo4j.NotificationDisabledCategory`
   * Stabilize GQL status objects (use this instead of notifications):
     * `ResultSummary.gql_status_objects`
     * `neo4j.GqlStatusObject`
 * Errors
   * Stabilize GQL error data
     * `neo4j.exceptions.GqlError`
     * `neo4j.exceptions.GqlErrorClassification`

PR #1240, opened 2025-09-12
Amend notification APIs deprecation in favor of GQL status APIs

robsdedude:
This PR fixes some docs that were missed in https://github.com/neo4j/neo4j-python-driver/pull/1209


AndyHeap-NeoTech (approved):
Everything looks correct.

PR #1241, opened 2025-09-12
ADR 033: UnsupportedType

robsdedude:
Bolt 6 introduces a new data type representing a value of a type that the server cannot transmit to the client due to the negotiation of an incompatible bolt protocol version.

To do:
 * https://github.com/neo4j-drivers/testkit/pull/688/
 * [x] decide which package the new type should live in
 * [x] finalize name of the new type (team discussion pending)


MaxAake (approved):
✅

PR #1251, opened 2025-09-29
API docs: fix server version needed for vectors

robsdedude:
At the time of writing (and very likely the next release), no server with non-experimental Bolt 6.0 support has been released. Therefore, this note needs to go for now.


StephenCathcart (approved):
👍

PR #1254, opened 2025-10-01
Add GitHub actions: windows unit tests

robsdedude:
* Adding GitHub Actions CI to run unit tests under Windows
* Work around pytest on windows not liking long test ids
  https://github.com/pytest-dev/pytest/issues/6881
* Fix test flakiness on systems with low monotonic clock resolution  
  👀️🪟️
* Fix DNS error re-write on Windows    
  On windows only, the driver would re-write the DNS error when resolving `Address(None, None)` to be retryable, which it shouldn't.


MaxAake (approved):
✅

PR #1255, opened 2025-10-20
Fix driver factory type hints suggesting unsupported options

robsdedude:
The type hints of the driver factory incorrectly suggested that the config options `max_transaction_retry_time` and `connection_acquisition_timeout` are unsupported.


AndyHeap-NeoTech (approved):
Good to go

PR #1258, opened 2025-10-21
[5.28] Add GitHub actions: windows unit tests

robsdedude:
* Adding GitHub Actions CI to run unit tests under Windows
* Work around pytest on windows not liking long test ids
* Fix old numpy and pandas versions treating `int` as `int32` on windows.

Back port of https://github.com/neo4j/neo4j-python-driver/pull/1254

PR #1261, opened 2025-10-24
Harden serialization of temporal numpy types

robsdedude:
Partial port of https://github.com/neo4j/neo4j-python-driver/pull/1258

The code isn't broken, but this way, things are more explicit and are less likely to develop unexpected behavior if things in numpy change in the future.


MaxAake (approved):
✅

PR #1263, opened 2025-11-05
Fix vector type native adapters

robsdedude:
`Vector.from_native` and `Vector.to_native` both used to have alternative implementations using NumPy or the Rust extensions when available to speed up the process. It turns out, that both didn't follow the reference Python implementation (internally relying on `struct.pack`/`struct.unpack`) for 32 bit float values.

 * Python 3.14+ preserves the signaling/quieting bit of `NaN` values whereas NumPy doesn't and Rust doesn't make strong guarantees on how f64s are cast to f32s: https://github.com/rust-lang/rfcs/blob/master/text/3514-float-semantics.md
 * Both NumPy and Rust turn large f64s into `inf` when cast to f32, while Python raises an `OverflowError`.

To harmonize the behavior, only the Python implementation remains. It has been significantly sped up by not iterating over the elements and calling `struct.(un)pack` multiple times, but instead making use of the fact that the format string allows to specify a number of values to (un)pack. With this the performance of the Python implementation was never slower than 5 times of what Rust/NumPy offered and for low dimension vectors even faster. It's deemed not worth the effort of maintaining multiple implementations assuring their behavioral parity for such a small performance gain.

The same multi-value (un)pack approach has been applied to the pure Python implementation of the byte swapping to improve it's performance by a factor of roughly 30 (still a factor of ~10 slower than using NumPy/Rust).

Fixes: DRIVERS-162

### DIFF
--- a/src/neo4j/vector.py
+++ b/src/neo4j/vector.py
@@ -43,11 +43,9 @@ if False:
 
 
 try:
-    from ._rust import vector as _vec_rust
     from ._rust.vector import swap_endian as _swap_endian_unchecked_rust
 except ImportError:
     _swap_endian_unchecked_rust = None
-    _vec_rust = None
 
 
 __all__ = [
@@ -602,11 +600,19 @@ def _swap_endian_unchecked_np(type_size: int, data: bytes, /) -> bytes:
 
 
 def _swap_endian_unchecked_py(type_size: int, data: bytes, /) -> bytes:
-    return bytes(
-        byte
-        for i in range(0, len(data), type_size)
-        for byte in data[i : i + type_size][::-1]
-    )
+    match type_size:
+        case 2:
+            fmt = "h"
+        case 4:
+            fmt = "i"
+        case 8:
+            fmt = "q"
+        case _:
+            raise ValueError(f"Unsupported type size: {type_size}")
+    count = len(data) // type_size
+    fmt_be = f">{count}{fmt}"
+    fmt_le = f"<{count}{fmt}"
+    return _struct.pack(fmt_be, *_struct.unpack(fmt_le, data))
 
 
 if _swap_endian_unchecked_rust is not None:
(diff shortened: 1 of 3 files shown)


src/neo4j/vector.py
```diff
@@ -602,11 +600,19 @@ def _swap_endian_unchecked_np(type_size: int, data: bytes, /) -> bytes:
...
+        case 4:
+            fmt = "i"
+        case 8:
+            fmt = "q"
+        case _:
+            raise ValueError(f"Unsupported type size: {type_size}")
+    count = len(data) // type_size
+    fmt_be = f">{count}{fmt}"
+    fmt_le = f"<{count}{fmt}"
+    return _struct.pack(fmt_be, *_struct.unpack(fmt_le, data))
```
RichardIrons-neo4j, line 615: This is clever stuff, I like it.

tests/unit/common/vector/test_vector.py
```diff
@@ -322,12 +334,12 @@ def test_swap_endian_unhandled_size(mocker, ext, type_size):
...
         pytest.param(
             "i32",
-            b"\x00\x00\x00\x01",
+            bytes.fromhex("00000001"),
             id="i32-single",
         ),
         pytest.param(
             "i32",
-            b"\x00\x00\x00\x01\x00\x00\x00\x02",
+            bytes.fromhex("0000000100000002"),
```
RichardIrons-neo4j, line 342: are you allowed spaces in these strings or anything?

tests/unit/common/vector/test_vector.py
```diff
@@ -1079,12 +1138,16 @@ def _dtype_to_cypher_type(dtype: T_DTYPE_LITERAL) -> str:
...
+def _vec_element_cypher_repr(value: t.Any, dtype: T_DTYPE_LITERAL) -> str:
+    if isinstance(value, float) and dtype in {"f32", "f64"}:
         if math.isnan(value):
             return "NaN"
         if math.isinf(value):
             return "Infinity" if value > 0 else "-Infinity"
+    if dtype == "f32":
+        # account for float32 precision loss
+        compressed = struct.unpack(">f", struct.pack(">f", value))[0]
+        return repr(compressed)
```
RichardIrons-neo4j, line 1150: is this that thing from the call the other day?

RichardIrons-neo4j (approved):
Good change, makes the code much easier to understand.

tests/unit/common/vector/test_vector.py
robsdedude: > [...] with ASCII whitespace being ignored.  
> -- https://docs.python.org/3.10/library/stdtypes.html#bytes.fromhex

Positive. The thought behind this change was to a) make the byte string more copiable and b) more compact.

tests/unit/common/vector/test_vector.py
robsdedude: Similar. If you take a very small f64 value and cast it to f32, it might end up being exactly 0.0 because f32s have fewer decimal (actually binary) places.

tests/unit/common/vector/test_vector.py
RichardIrons-neo4j: @robsdedude I would put some spaces in then. Not going to hold up the PR over it though.

PR #1267, opened 2025-11-27
Fix AttributeError when pool is None in Bolt3 error handling

kirill-stepanishin:
### Problem
When a Neo4jError is raised in Bolt3 connections, the error handler attempts to call self.pool.on_neo4j_error() without checking if self.pool exists. This causes an AttributeError when the connection pool is None or not available.

### Solution
Added a null check for self.pool before calling on_neo4j_error(), consistent with the existing error handling pattern used for ServiceUnavailable, DatabaseUnavailable, NotALeader, and ForbiddenOnReadOnlyDatabase exceptions in the same method.

### Changes
- src/neo4j/_async/io/_bolt3.py: Added if self.pool: check before await self.pool.on_neo4j_error()
- src/neo4j/_sync/io/_bolt3.py: Auto-generated sync version with the same fix

This prevents crashes when handling Neo4j errors in scenarios where the connection pool is not initialized or has been cleared.

*Note:*
I am a first-time contributor and have just signed the CLA

### DIFF
--- a/src/neo4j/_async/io/_bolt3.py
+++ b/src/neo4j/_async/io/_bolt3.py
@@ -588,7 +588,8 @@ class AsyncBolt3(AsyncBolt):
                     )
                 raise
             except Neo4jError as e:
-                await self.pool.on_neo4j_error(e, self)
+                if self.pool:
+                    await self.pool.on_neo4j_error(e, self)
                 raise
         else:
             sig_int = ord(summary_signature)
--- a/src/neo4j/_sync/io/_bolt3.py
+++ b/src/neo4j/_sync/io/_bolt3.py
@@ -588,7 +588,8 @@ class Bolt3(Bolt):
                     )
                 raise
             except Neo4jError as e:
-                self.pool.on_neo4j_error(e, self)
+                if self.pool:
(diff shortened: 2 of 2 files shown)


robsdedude:
Thanks for taking the time to contribute :bow: This is much appreciated.

Just to set expectations: bolt 3 support in driver versions 5.0+ is not an official feature and is not tested in CI. We just maintain in on a best-efforts-it-doesnt-hurt-to-not-rip-it-out basis ;) this entails that we might remove it even in a minor version update if it becomes unreasonably complex to maintain.

I'm curious: did you actually come across a scenario in which the ´AttributeError´ was raised or is this an issue you found looking at the code?

For reference: issue introduced in https://github.com/neo4j/neo4j-python-driver/pull/890 (so this will need back porting to 5.0 once merged)

kirill-stepanishin:
Thank you for your response!

I ran into this while using Bolt v3 against Neptune. The auth errors it returns land in the generic Neo4jError handler, and that’s where I hit the AttributeError because the pool wasn’t set. The fix just let the original error bubble up properly.

robsdedude (approved):

PR #1276, opened 2025-12-11
Unify connection error handling between async and sync

robsdedude:
Error handling (especially logging) in the code establishing TCP & TLS connections differed too much making reasoning about it more complicated than it needs to be.

Spotted while working on: DRIVERS-233
Closes: DRIVERS-236


AndyHeap-NeoTech (approved):
Changes look good.

PR #1284, opened 2026-02-02
`Duration.from_iso_format`: allow negative components.

robsdedude:
`Duration.iso_format` could produce values that wouldn't be accepted by its counterpart `Duration.from_iso_format`: namely those with negative components.

E.g.,
```python
from neo4j.time import Duration

print(Duration(seconds=-1).iso_format())  # "PT-1S"
print(Duration.from_iso_format("PT-1S"))  # raised ValueError
```

Further, this PR:
 * Fixes previously rejected ISO formats containing weeks. E.g., `P1W` would raise `ValueError`.
 * Fixes invalid strings being accepted (e.g., `P` and `PT`).

Closes: DRIVERS-253


src/neo4j/time/__init__.py
```diff
@@ -685,10 +686,13 @@ def from_iso_format(cls, s: str) -> Duration:
             'P', a zero length duration
             'PyY', y being a number of years
             'PmM', m being a number of months
+            'PdW', d being a number of weeks
```
AndyHeap-NeoTech: Should this be PwW?

src/neo4j/time/__init__.py
```diff
@@ -699,23 +703,30 @@ def from_iso_format(cls, s: str) -> Duration:
...
             Any combination of all options, e.g. 'P13MT100M' for 13 months and
             100 minutes.
+            Except for weeks, which cannot be combined with any other
+            components.
 
         :param s: String to parse
 
         :raises ValueError: if the string does not match the required format.
         """
         match = _DURATION_ISO_PATTERN.match(s)
```
AndyHeap-NeoTech, line 722: Just an observation coming in to this cold. The magic numbers for the groupings in the duration made me pause. Named constants would help with readability on first glance, but I can see how they could be considered superfluous when the values being assigned to are named.

src/neo4j/time/__init__.py
```diff
@@ -699,23 +703,30 @@ def from_iso_format(cls, s: str) -> Duration:
...
         if match:
+            if match.group(16):
```
AndyHeap-NeoTech: Just want to confirm weeks are never grouped with other time elements? So you can't have P5WT2H?

src/neo4j/time/__init__.py
```diff
@@ -699,23 +703,30 @@ def from_iso_format(cls, s: str) -> Duration:
...
+                return cls(weeks=int(match.group(16)))
             ns = 0
             if match.group(15):
                 ns = int(match.group(15)[1:10].ljust(9, "0"))
+            seconds = int(match.group(14) or 0)
```
AndyHeap-NeoTech: Nanoseconds are the only component that need to have their sign changed if another is negative? E.g. It would not apply in the case of negative months meaning you need to negate days.

src/neo4j/time/__init__.py
robsdedude: I think constants might make it harder to read as it increases clutter while, as you said, the semantics of the magic number is right next to in in the name of the variable/function argument it's being assigned to/used for.

I'll try using named regex groups to see how that feels and looks like.

robsdedude: Looks better, I'd say. And performance seems to be on par as well. :+1:

src/neo4j/time/__init__.py
robsdedude: That's right. ISO 8601 says either weeks or a (non-empty) combination of the others. Now Java's default ISO duration parser appears to be more liberal and allows for mixing weeks with other components. However, I decided to stick closer to the standard (at least for now).

src/neo4j/time/__init__.py
robsdedude: If you take a look at the regex, all groups but nanoseconds are `[+-]?\d+`. I.e., they permit a sign. Only nanoseconds doesn't because it the after-comma digits of seconds, i.e., the second group in `([+-]?\d+)([.,]\d+)`. Therefore it needs some special handling when parsing (among copying the sign of seconds).

AndyHeap-NeoTech (approved):
Looks good.

PR #1285, opened 2026-02-03
Improve error message on invalid `Vector` init overload invocation

robsdedude:
Consider the following (erroneous) example driver usage

```python
from neo4j.vector import Vector
import numpy as np


np_array = np.array([1, 2, 3], dtype=np.int64)
Vector(np_array, "i64")
```
The issue here is that when a numpy array is being used to construct a vector, no dtype argument is accepted because the numpy array already carries the type information.

The result error, however, was arguably confusing and little helpful. It might even appear to be a driver-internal issue:
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "~/tmp/example.py", line 6, in <module>
    Vector(np_array, "i64")
  File ".../neo4j/vector.py", line 181, in __init__
    self._set_numpy(data, *args, **kwargs)
TypeError: Vector._set_numpy() takes 2 positional arguments but 3 were given
```

This PR aims to improve the error and hopes to guide developers to the underlying issue quicker:
```
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "~/tmp/example.py", line 6, in <module>
    Vector(np_array, "i64")
  File ".../neo4j/vector.py", line 186, in __init__
    with self._checked_init_overload(
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File ".../contextlib.py", line 158, in __exit__
    self.gen.throw(value)
  File ".../neo4j/vector.py", line 224, in _checked_init_overload
    raise TypeError(
TypeError: Invalid arguments for Vector initialization with data of type numpy.ndarray: too many positional arguments
```

Closes: DRIVERS-186


src/neo4j/vector.py
```diff
@@ -176,13 +178,54 @@ def __init__(self, data: _pa.Array, /) -> None: ...
...
+        args: tuple,
+        kwargs: dict[str, _t.Any],
+    ) -> _t.Generator[None, None, None]:
+        try:
+            yield
+        except TypeError:
+            signature = _inspect.signature(setter)
+            signature = signature.replace(
+                return_annotation=_inspect.Signature.empty
+            )
```
MaxAake: This appears to be unused code.

MaxAake (changes requested):

MaxAake (approved):
✅

PR #1291, opened 2026-02-06
Pandas 3 support

robsdedude:
Adding support for [`pandas`](https://pandas.pydata.org/) 3.x.

The driver might indirectly behave differently when used with pandas 3. For example, calling `Result.to_df` on a `Result` containing a string-only column produces a `pandas.DataFrame` with `dtype` `object` for this column. When using `pandas` 3, the `dtype` will be `StringDtype` instead.

Closes: DRIVERS-250

### DIFF
--- a/tests/unit/async_/work/test_result.py
+++ b/tests/unit/async_/work/test_result.py
@@ -74,6 +74,10 @@ if t.TYPE_CHECKING:
     )
 
 
+# https://pandas.pydata.org/docs/user_guide/migration-3-strings.html
+PD_STR_DTYPE = "object" if pd.__version__ < "3" else "string"
+
+
 class Records:
     def __init__(self, fields, records):
         self.fields = tuple(fields)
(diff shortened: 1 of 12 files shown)


tests/unit/async_/work/test_result.py
```diff
@@ -74,6 +74,10 @@
     )
 
 
+# https://pandas.pydata.org/docs/user_guide/migration-3-strings.html
+PD_STR_DTYPE = "object" if pd.__version__ < "3" else "string"
```
AndyHeap-NeoTech: Just want to confirm that this evaluates to a constant rather than running the conditional logic every time it is used.

AndyHeap-NeoTech (approved):

tests/unit/async_/work/test_result.py
robsdedude: Think of it more like a static than a constant. It's a global variable and nothing but convention (it being all-caps) stops code from mutating its value. But yes, the condition is only evaluated once.

PR #1292, opened 2026-02-06
Fix Python 3.10 dev env

robsdedude:
Sphinx dropped support for some Python versions we still want to support. In CI we'll still use the latest Python version the driver supports to build the docs. Therefore, we can be sure we're building our docs with the latest available version of Sphinx.

Closes: DRIVERS-262


AndyHeap-NeoTech (approved):
LGTM

PR #1295, opened 2026-02-27
Speed up record iteration

robsdedude:
For the query `UNWIND range(1, 10_000) AS n RETURN n` using `driver.execute_query`, this change yields a performance boost of roughly 10%.

Closes: DRIVERS-277

### DIFF
--- a/src/neo4j/_async/work/result.py
+++ b/src/neo4j/_async/work/result.py
@@ -249,8 +249,7 @@ class AsyncResult(AsyncNonConcurrentMethodChecker):
                     for record in records
                 )
                 self._record_buffer.extend(
-                    Record(zip(self._keys, record, strict=True))
-                    for record in records
+                    Record._new(self._keys, record) for record in records
                 )
 
         async def _on_summary():
@@ -459,10 +458,9 @@ class AsyncResult(AsyncNonConcurrentMethodChecker):
             record_buffer.append(record)
             if n is not None and len(record_buffer) >= n:
                 break
-        if n is None:
-            self._record_buffer = record_buffer
-        else:
-            self._record_buffer.extend(record_buffer)
+        if self._record_buffer:
+            record_buffer.extend(self._record_buffer)
+        self._record_buffer = record_buffer
         self._exhausted = not self._record_buffer
 
     async def _buffer_all(self):
--- a/src/neo4j/_data.py
+++ b/src/neo4j/_data.py
@@ -54,7 +54,7 @@ class Record(tuple, t.Mapping):
     values rather than keys.
     """
 
-    __keys: tuple[str]
+    __keys: tuple[str, ...]
 
     def __new__(cls, iterable=()):
         keys = []
@@ -62,6 +62,17 @@ class Record(tuple, t.Mapping):
         for key, value in iter_items(iterable):
             keys.append(key)
             values.append(value)
+        inst = tuple.__new__(cls, values)
+        inst.__keys = tuple(keys)
+        return inst
+
+    @classmethod
+    def _new(
+        cls,
+        keys: t.Collection[str],
+        values: t.Collection[t.Any],
+    ) -> t.Self:
+        assert len(keys) == len(values)
         inst = tuple.__new__(cls, values)
         inst.__keys = tuple(keys)
         return inst
--- a/src/neo4j/_data.py
+++ b/src/neo4j/_data.py
@@ -72,7 +72,12 @@ class Record(tuple, t.Mapping):
         keys: t.Collection[str],
         values: t.Collection[t.Any],
     ) -> t.Self:
-        assert len(keys) == len(values)
+        if len(keys) != len(values):
+            raise ValueError(
+                f"keys and values have different length ({keys=}, {values=}); "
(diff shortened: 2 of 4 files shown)


src/neo4j/_async/work/result.py
```diff
@@ -249,8 +249,7 @@ def on_records(records):
                     for record in records
                 )
                 self._record_buffer.extend(
-                    Record(zip(self._keys, record, strict=True))
-                    for record in records
+                    Record._new(self._keys, record) for record in records
```
AndyHeap-NeoTech, line 252: Just want to ensure I understand the optimisation fully. Essentially it is removing the creation of an intermediate zip object before passing into _new.

Does _new still throw an error if the two are different lengths (as zip does with strict=True)?

robsdedude: > Just want to ensure I understand the optimisation fully

This saves time by avoiding

  - performing a zip for each record
  - creating two lists per record https://github.com/neo4j/neo4j-python-driver/pull/1295/changes#diff-de3b56aa7dd7eb38fd0c6d49edc75abf83593ac4e6469c8edb440036f8c2feecR60-R61
   - calling [`iter_items`](https://github.com/robsdedude/neo4j-python-driver/blob/665508ee7f7dc61d4dea11f35b25f3fe68a26baf/src/neo4j/_conf.py#L33-L46) which creates a generator and then iterating over that generator
   - incrementally populating these lists https://github.com/neo4j/neo4j-python-driver/pull/1295/changes#diff-de3b56aa7dd7eb38fd0c6d49edc75abf83593ac4e6469c8edb440036f8c2feecR63-R64
   - then converting the lists to tuples

All of that is being done to arrive back at exactly the tuples that were already available as `self._keys` and `record`.

> Does _new still throw an error if the two are different lengths (as zip does with strict=True)?

It does throw unless users disable [assertions](https://docs.python.org/3/reference/simple_stmts.html#assert). See https://github.com/neo4j/neo4j-python-driver/pull/1295/changes#diff-de3b56aa7dd7eb38fd0c6d49edc75abf83593ac4e6469c8edb440036f8c2feecR75

Granted, this should never happen and would either be a protocol error or a programming error when interpreting a `RECORD` message. For the latter, I'd be fine with the assertion being optional. For the former, not so much. Maybe I should turn this into a check that cannot be disabled, wdyt?

src/neo4j/_async/work/result.py
```diff
@@ -459,10 +458,9 @@ async def _buffer(self, n=None):
             record_buffer.append(record)
             if n is not None and len(record_buffer) >= n:
                 break
-        if n is None:
-            self._record_buffer = record_buffer
-        else:
-            self._record_buffer.extend(record_buffer)
+        if self._record_buffer:
```
AndyHeap-NeoTech, line 461: This changes the ordering in the buffer. Original: new records go on the end. PR: new records go at the front. This is a behavioural change, is it intentional?

I'm wondering if this breaks anything and whether there is an existing test that might catch this. Don't think so as it would have blown up with the change.

src/neo4j/_async/work/result.py
AndyHeap-NeoTech: It is probably an unnecessary check to do at this point. Shouldn't a protocol error to have arisen before this if the data is this malformed?

AndyHeap-NeoTech (approved):
Core change is very nice. Just a question around ordering in a buffer.

src/neo4j/_async/work/result.py
robsdedude: I'm not aware of any other part in the driver checking that the number of fields of each record match the number of announced fields in the `RUN`'s `SUCCESS` response.

src/neo4j/_async/work/result.py
robsdedude: Thanks good catch. I'm actually fairly certain that this condition is never true anyway, such that the `if`'s body is never executed.

Either way, the order would've been messed up by the code as it was before should the condition ever be true, so I argue this PR fixes the code, just that the code is never executed.

Consider the old code:

 * Assume the for loop in `self._buffer(...)` ends up yielding only one record, hence `record_buffer = [r1]`
 * The buffer is not empty, it contains `self._record_buffer = [r2, r3]` (note, those records *must* be later than `r1` assuming that iterating over the result correctly yield records in order)
 * now the old logic `self._record_buffer.extend(record_buffer)` yield `self._record_buffer = [r2, r3, r1]` ... 😱

AndyHeap-NeoTech: I agree, it's a bug fix, even if the condition never triggers :) 

If it never triggers is there a reason to have it? Or is it a case of bullet proofing things _just in case_?

robsdedude: The latter. The invariant that the buffer is empty after iterating for at least the buffer's size many records is not documented, spread across many places, and untested. Besides, it might be dropped accidental or by design in the future (e.g., introducing a multi record bolt response might lead to this). Therefore, I'd like this code to be ready to handle it.

RichardIrons-neo4j (approved):
Looks good. Good catch from Andy on the buffer reordering.

PR #1300, opened 2026-04-08
ADR 038: idempotent retries for session.run

robsdedude:
This PR introduces one-shot retries without delays to `session.run`. Retries on `Session.run()` are limited to a specific set of errors. Namely those errors marked as *idempotent* by the DBMS, i.e., errors that are guaranteed to not have altered the state of any database. Currently, this set encompasses only on admission control errors.

Further, a new configuration option at driver *and* session level called `disable_auto_commit_retries` is being added to allow users to opt out of this new behavior. However, we expect the impact to be negligible for most use-cases.

Closes: DRIVERS-310

## Open TODOs:

  * [x] Wait for ADR to be merged
    * [x] Copy and adjust docs suggestions for `disable_auto_commit_retries`
    * [x] One final pass of the ADR to make sure the PR adheres to the specs
  * [x] Wait for TestKit to be ready
    * [x] Add TestKit feature
    * https://github.com/neo4j-drivers/testkit/pull/708
    * [x] Run CI to test against TestKit


src/neo4j/_sync/work/session.py
```diff
@@ -301,37 +301,66 @@ def run(
...
+                    self._config.notifications_disabled_classifications,
+                )
+            except Neo4jError as error:
+                if (
+                    self._config.disable_auto_commit_retries
+                    or not auto_result._attach_failed
+                    or retry_count > 0
+                    or error.diagnostic_record.get("_idempotent") is not True
+                ):
+                    raise
```
RichardIrons-neo4j, line 355: This is such a tiny thing just so you know I really am reading the code - I think I'd have used `retried = true` for this, _or_ if the consideration is "we might allow more retries in future", I'd make a constant somewhere like max_idempotent_retries and reference that.

RichardIrons-neo4j (approved):

src/neo4j/_sync/work/session.py
robsdedude: I'll be going with a constant then, although if we ever increase the max number of retries, we'll probably also introduce some waiting logic, so this piece of code will likely have to be touched anyway.

RichardIrons-neo4j: 👍

RichardIrons-neo4j (approved):

PR #1301, opened 2026-04-21
Fix socket timeouts during connection establishment

robsdedude:
* Prevent async socket from passing 0 as `ssl_handshake_timeout` leading to
    a `ValueError` being raised when the acquisition deadline has expired.
    Raise `ServiceUnavailable` instead.
  * Align async driver with sync driver:
    * (lazily) raise a `ValueError` when configured with a negative
      `connection_timeout`.
    * Treat `connection_timeout=0` as "no timeout".
  * Fix sync driver setting the socket timeout to `0` (non-blocking) on expired
    `connection_acquisition_timeout`.
  * Harden internal code against against `None` timeouts.
  
Closes: DRIVERS-331

Amends: https://github.com/neo4j/neo4j-python-driver/pull/1274

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,12 +110,14 @@ test = [
     "pytest-benchmark >= 5.1.0",
     "pytest-cov >= 7.0.0",
     "pytest-mock >= 3.15.1",
+    "mock >= 5.2.0"
 ]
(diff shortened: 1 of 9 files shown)


RichardIrons-neo4j (approved):
Nice job 🍄

PR #1307, opened 2026-05-05
ADR 036: add benchmarking workloads

robsdedude:
Closes: DRIVERS-354

### DIFF
--- a/benchmarks/src/bench/workloads/api_unmanaged_transaction.py
+++ b/benchmarks/src/bench/workloads/api_unmanaged_transaction.py
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # * *date*: 2025-10-27
-# * *id*: `api-session-run`
+# * *id*: `api-unmanaged-transaction`
 
 
 from __future__ import annotations
--- a/benchmarks/src/bench/workloads/api_unmanaged_transaction.py
+++ b/benchmarks/src/bench/workloads/api_unmanaged_transaction.py
@@ -1,0 +1,49 @@
+# Copyright (c) "Neo4j"
+# Neo4j Sweden AB [https://neo4j.com]
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# * *date*: 2025-10-27
+# * *id*: `api-session-run`
+
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+from .util import benchmark
+
+
+if TYPE_CHECKING:
+    from neo4j import Driver
+
+    from .util import Bencher
+
+
+@benchmark
+def api_unmanaged_transaction(bencher: Bencher) -> None:
+    with bencher.ctx.new_driver() as driver:
+        bencher.loop(_work, bencher, driver)
+
+
+def _work(bencher: Bencher, driver: Driver) -> None:
+    with (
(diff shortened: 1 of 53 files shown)


benchmarks/src/bench/workloads/util/_timer.py
```diff
@@ -0,0 +1,81 @@
...
+        self._runs += 1
+
+    def reset(self) -> None:
+        if self._start_time is not None:
+            raise RuntimeError("Cannot reset timer while it is running")
+        self._runs = 0
+        self._timings = []
+
+    def flush_csv(self, file_path: Path) -> None:
+        if self._n != len(self._timings):
```
AndyHeap-NeoTech: Might be a bug here. _n is the number of runs, but the length of _timings will be n - warmup (see if clause on line 56).

benchmarks/src/bench/workloads/util/_timer.py
```diff
@@ -0,0 +1,81 @@
...
+            raise RuntimeError(
+                f"Expected {self._n} timings, but got {len(self._timings)}"
+            )
+        with file_path.open("w", newline="", encoding="utf-8") as csvfile:
+            writer = csv.writer(csvfile)
+            for timing in self._timings:
+                writer.writerow([timing.start, timing.start, timing.end])
```
AndyHeap-NeoTech: Should it be writing out timing.start twice? Perhaps it's meant to be start, end and duration or something?

benchmarks/src/bench/workloads/util/_benchmark_registry.py
```diff
@@ -0,0 +1,72 @@
...
+        attempt_left = 30
+        exceptions: list[Exception] = []
+        while attempt_left > 0:
+            try:
+                func(bencher)
+            except Exception as e:
+                exceptions.append(e)
+            else:
+                return
+            attempt_left -= 1
```
robsdedude, line 60: Needs double checking: is Bencher appropriately reset between retries?

benchmarks/src/bench/workloads/api_unmanaged_transaction.py
```diff
@@ -0,0 +1,49 @@
...
+    with bencher.ctx.new_driver() as driver:
+        bencher.loop(_work, bencher, driver)
+
+
+def _work(bencher: Bencher, driver: Driver) -> None:
+    with (
+        driver.session(database="system") as session,
+        bencher.timer,
+    ):
+        for _ in range(2):
```
AndyHeap-NeoTech, line 44: Just wondering why the magic number 2. Why two iterations on this?

AndyHeap-NeoTech (commented):
Looks good, a possible bug to sanity check and one comment.

benchmarks/src/bench/workloads/util/_timer.py
robsdedude: N in timer is actually only the number of iterations. I've renamed all occurrences of `n` to be explicit whether they're `n_total` or `n_timed`. I realized that's really needed here as I had to carefully track the flow myself to figure out whether this was the right "n" or not.

benchmarks/src/bench/workloads/util/_timer.py
robsdedude: The CSV columns expected by the framework are

* `scheduled_at` value is an epoch millis points to the instant of the scheduled measurement,
* `started_at` value is an epoch millis points to the instant of the measurement start time,
* `stopped_at` value is an epoch millis points to the instant of the measurement finish time.

Since we don't really have any scheduling of work, I figured I'll just treat the start time as the time the work has been scheduled.

benchmarks/src/bench/workloads/api_unmanaged_transaction.py
robsdedude: Because the workload description in the ADR says so.

Now... Granted, I wrote that ADR, too :clown_face: But I'd prefer to have that discussion over there as it applies to all drivers alike.

AndyHeap-NeoTech: Ok makes sense. I'll add a comment to the ADR.

AndyHeap-NeoTech (approved):
Looks good. Thanks for the clarifications.

PR #1308, opened 2026-05-11
Bolt 6.1: UUID type

robsdedude:
Add support for Bolt 6.1 which includes a new UIDD type which maps to Python's builtin [`uuid.UUID`](https://docs.python.org/3.10/library/uuid.html#uuid.UUID) type. N.B., `UUID.is_safe` will not be preserved when round-tripped through the DBMS.

Closes: DRIVERS-365

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/713


src/neo4j/_codec/packstream/v2/__init__.py
```diff
@@ -0,0 +1,270 @@
...
+PACKED_UINT_8 = v1.PACKED_UINT_8
+PACKED_UINT_16 = v1.PACKED_UINT_16
+
+UNPACKED_UINT_8 = v1.UNPACKED_UINT_8
+UNPACKED_UINT_16 = v1.UNPACKED_UINT_16
+
+INT64_MIN = v1.INT64_MIN
+INT64_MAX = v1.INT64_MAX
+
+
```
RichardIrons-neo4j, line 59: These classes seem to have a lot of duplicate code with the v1 versions. Could it have something like `if isinstance(value, UUID): doUuidStuff(...) else: super()._py_pack(...)`?

RichardIrons-neo4j:
Also note: should CHANGELOG.md have a line under 6.1 for this?

RichardIrons-neo4j (changes requested):

src/neo4j/_codec/packstream/v2/__init__.py
robsdedude: It could. I deliberately decided against it for some reasons:
 * This it the hot path of the driver. It's one of the bits of code that the rust extensions replace (as suggested by the thing being called `_py_pack` as opposed to `_rust_pack`). So an extra function call for each value to pack as well as each of its inner values has very likely a high performance impact.
 * I've not tried to implement it yet, but I think one couldn't just descend into `super()._py_pack()` for a recursive structure like a list or map, without some way to inject the new packer logic.
 * This function is recursive over the value to be serialized. Tbf, it probably shouldn't be recursive in the first place, but here we are. Now calling `super` would double (?) the number of required stack frames. Python applies a rather low limit by default in the order of 1k. So I do think that'd be rather limiting for users.

RichardIrons-neo4j: Ah yeah, the recursion thing especially could get really awkward. Approved!
:lgtm:
(I can't believe that's not an emoji).

RichardIrons-neo4j (approved):
🥳

PR #1311, opened 2026-06-02
Fix async driver waiting for OS on closing half-open socket

### LINKED ISSUE #1309: Async pool: acquire blocks far past connection_acquisition_timeout on unbounded teardown of a discarded half-open connection
## Summary

In the **async** pool, when a request is handed a connection that turns out to be unusable (stale, or failing its liveness check), the driver discards it by calling `await connection.close()` **synchronously inside the acquisition path**. The async `close()` awaits the full socket teardown — `await writer.wait_closed()`. If that connection's socket is *half-open* — the peer is gone but no `RST`/`FIN` was received (a common outcome when an idle TCP connection is silently dropped by an intervening load balancer, NAT, or firewall) — then `wait_closed()` blocks on the OS TCP/TLS teardown, which can take tens of seconds.

During that time the acquiring coroutine makes no progress. The eventual error is:

```
neo4j.exceptions.ConnectionAcquisitionTimeoutError: failed to obtain a connection from the pool within 10.0s (timeout)
```

…but the wall-clock time spent is **far longer than the configured `connection_acquisition_timeout`** (e.g. ~30s observed with a 10s timeout). The `"within 10.0s"` text is just the configured value formatted into the message; the deadline was not actually enforced, because the blocking step is the connection *teardown*, which is not covered by the acquisition deadline.

The CHANGELOG for 6.0 states the option “now spans *anything* required to acquire a connection.” Discarding an unusable connection is part of acquiring one, so an unbounded teardown inside that path appears to violate that contract.

> Note: the blocking step is the teardown (`wait_closed()`), **not** the `GOODBYE` write. The `GOODBYE` send is small, buffered, and its errors are swallowed; it returns promptly. The stall is in the subsequent `await self.socket.close()`.

## Environment

- `neo4j` driver **6.2.0**, **async** driver. (The async pool is the affected one; the same `await ... wait_closed()` pattern is present in earlier 5.x/6.x async releases.)
- The **sync** pool is **not** affected — see the note at the end.
- TLS connection (`neo4j+s://`) to a managed cluster behind a load balancer; connections go idle in the pool and are silently dropped by intervening network equipment (the driver docs themselves note this can happen).

## Where it happens

The teardown that blocks is `AsyncBoltSocket.close()` (`src/neo4j/_async_compat/network/_bolt_socket.py`):

```python
async def close(self):
    self._writer.close()
    await self._writer.wait_closed()   # blocks on TLS/TCP teardown of a half-open socket
```

There are **two** places during acquisition that reach this unbounded teardown:

**1. Stale connection** — `AsyncIOPool._acquire_from_pool_checked` (`src/neo4j/_async/io/_pool.py`, ~L199–L227):

```python
async def _acquire_from_pool_checked(self, address, health_check, deadline):
    while not deadline.expired():
        connection = await self._acquire_from_pool(address)
        if not connection:
            return None
        if not await health_check(connection, deadline):   # stale -> False immediately, no probe
            ...  # "found unhealthy connection ..."
            await connection.close()        # <-- L222: unbounded teardown
            self._remove_connection(connection)
            continue
        else:
            return connection
    return None
```

For a stale connection, `health_check` returns `False` *immediately* (no liveness probe), so the hang is the `await connection.close()` at L222.

**2. Failed liveness check** — the probe itself *is* bounded by the acquisition deadline via `connection_deadline(...)`, **but** when it fails on a dead socket the driver routes through `AsyncBolt._set_defunct` (`src/neo4j/_async/io/_bolt.py`), which itself awaits a close (~L950):

```python
self._defunct = True
...
if not self._closing:
    await self.close()          # <-- L950: same unbounded teardown (GOODBYE skipped since defunct)
    if (not self._liveness_check and self.pool and ...):
        await self.pool.deactivate(address=self.unresolved_address)
```

Crucially, `connection_deadline(...)` only sets the connection’s **read/write deadlines** (consulted by the `_wait_for_read` / `_wait_for_write` wrappers). It does **not** bound `wait_closed()`, so this teardown is unbounded as well. This hang occurs inside `health_check` (called at `_pool.py` ~L206); by the time L222 runs the connection is already `_closing`, so the close there is a no-op. So the L222 site alone does not cover this path.

The driver already has a non-blocking teardown — `AsyncBoltSocket.kill()` (`self._writer.close()` with **no** `await wait_closed()`; idempotent, sets `_closed`) — but the discard paths use the graceful `close()`.

## Reproduction

The real-world trigger is an idle pooled connection whose socket is silently dropped by a load balancer/NAT, then reused. A deterministic stand-in makes the socket teardown hang and forces the pooled connection to be stale before the next acquire:

```python
import asyncio
from neo4j import AsyncGraphDatabase

async def main():
    driver = AsyncGraphDatabase.driver(
        "neo4j+s://db.example.com:7687",
        auth=("neo4j", "password"),
        max_connection_lifetime=1,          # make connections go stale quickly
        connection_acquisition_timeout=5,   # we EXPECT the next acquire to fail in ~5s
    )

    # 1) Populate the pool with one connection.
    async with driver.session() as s:
        await (await s.run("RETURN 1")).consume()

    # 2) Simulate a half-open socket: make this connection's teardown hang,
    #    as it would if a load balancer silently dropped the idle TCP flow.
    conn = next(iter(driver._pool.connections.values()))[0]
    async def hanging_close():
        await asyncio.sleep(60)             # stand-in for the OS TCP/TLS teardown timeout
    conn.socket.close = hanging_close

    # 3) Let the pooled connection age past max_connection_lifetime (-> stale).
    await asyncio.sleep(2)

    # 4) Acquire again. The pool finds the stale connection and awaits its
    #    close() -> blocks ~60s, far past connection_acquisition_timeout=5.
    loop = asyncio.get_event_loop()
    t0 = loop.time()
    try:
        async with driver.session() as s:
            await (await s.run("RETURN 1")).consume()
    except Exception as e:
        print(f"elapsed={loop.time() - t0:.1f}s  ->  {e!r}")

    await driver.close()

asyncio.run(main())
```

**Observed:** `elapsed≈60.0s -> ConnectionAcquisitionTimeoutError(... within 5.0s ...)`
**Expected:** the acquire should not block on tearing down a connection it has already decided to discard; it should obtain/create a healthy connection within `connection_acquisition_timeout`.

(The same hang is reachable via the failed-liveness path through `_set_defunct` when the connection is idle-but-not-stale. Without the monkeypatch, the behavior also reproduces against a real server by dropping packets to the server port — e.g. an `iptables`/firewall `DROP` rule — while a stale connection sits in the pool, then issuing a query.)

Representative DEBUG log shape (placeholder IDs), showing the gap between detecting the unhealthy connection and the eventual timeout, with nothing in between:

```
... <POOL> acquire address, database='neo4j' address=<host>:7687
... [#A1B2] <POOL> found unhealthy connection bolt-00000 (closed=False, defunct=False, stale=True, in_use=True)
... [#A1B2] C: GOODBYE
... [#A1B2] C: <CLOSE>
        << ~30s with no activity — blocked in socket teardown (wait_closed) >>
... [#A1B2] <POOL> remove connection from pool <host>:7687 bolt-00000
... [#0000] <POOL> acquisition timed out
```

## Proposed fix (in words)

A connection the driver has already decided to **abandon** (stale, failed liveness, or defunct) does not need a graceful `GOODBYE` — and on a half-open socket the `GOODBYE` can’t be delivered anyway. So acquisition should tear such connections down with the existing **non-blocking** `kill()` (`writer.close()` without `await wait_closed()`) instead of the graceful `close()`. This must be applied at **both** sites above — the acquire-loop discard *and* `_set_defunct` — otherwise the failed-liveness path still hangs.

This also makes the async pool match the **sync** pool, whose `close()` is already a fire-and-forget `shutdown()` + `socket.close()` (no `wait_closed()` equivalent).

If preserving the graceful `GOODBYE` for stale-but-possibly-still-alive connections is desired, the alternative is to keep `close()` but move it **off the acquire path** (schedule it as a task, not awaited) and **bound** it (`asyncio.wait_for(...)` with a `kill()` fallback) — again at both sites.

## Proposed fix (code)

**Primary — `kill()` abandoned connections (non-blocking), at both sites:**

```python
# Site 1 — AsyncIOPool._acquire_from_pool_checked (_async/io/_pool.py)
if not await health_check(connection, deadline):
    if log.isEnabledFor(logging.DEBUG):
        log.debug(
            "[#%04X]  _: <POOL> found unhealthy connection %s",
            connection.local_port, connection.connection_id,
        )
    self._remove_connection(connection)
    connection.kill()        # non-blocking abort; was: await connection.close()
    continue

# Site 2 — AsyncBolt._set_defunct (_async/io/_bolt.py), connection-failed branch
if not self._closing:
    self.kill()              # non-blocking; was: await self.close()
    if (
        not self._liveness_check
        and self.pool
        and not self._get_server_state_manager().failed()
    ):
        await self.pool.deactivate(address=self.unresolved_address)
```

`kill()` is idempotent and already sets `_closing`/`_closed`, so the pool won’t reuse the connection and any later `close()` is a no-op. (The `if not connection_failed:` branch a few lines above in `_set_defunct` also does an unbounded `await self.close()` for the “unknown state” case; the same treatment applies there if that path is reachable on a half-open socket.)

**Alternative — keep graceful close but background + bound it** (apply at both sites):

```python
# init in AsyncIOPool.__init__:
#     self._discard_tasks: set[asyncio.Task] = set()
#     self._discard_close_timeout: float = 2.0

def _discard_connection(self, connection):
    async def _close_bounded():
        try:
            await asyncio.wait_for(connection.close(), self._discard_close_timeout)
        except (asyncio.TimeoutError, OSError, BoltError, DriverError):
            with suppress(OSError):
                connection.kill()
    task = asyncio.ensure_future(_close_bounded())
    self._discard_tasks.add(task)
    task.add_done_callback(self._discard_tasks.discard)
```

`AsyncIOPool.close()` should then cancel/await any pending `self._discard_tasks`.

## Note on the sync pool

The **sync** pool is **not** affected. Its `BoltSocket.close()` is `close_socket(...)` → `shutdown(SHUT_RDWR)` + `socket.close()` — a non-blocking, fire-and-forget teardown with no `wait_closed()` equivalent, so it does not block on a half-open socket (the OS finishes the FIN/teardown in the background; `SSLSocket.close()` does not perform a blocking SSL shutdown, and the driver sets no `SO_LINGER`). The bug is specific to the async driver’s `await writer.wait_closed()`. (A sync *liveness probe*’s `recv` on a dead socket can still block the calling thread, but that is bounded by the socket read timeout and is a separate path from the discard `close()`.)

Happy to open a PR if the maintainers agree on the direction (`kill()`-on-abandon vs. background-and-bound).

robsdedude:
This change aligns the async driver better with the sync driver. When a socket is closed, we don't want to wait until the OS acknowledges that the channel has been fully torn down. It's sufficient to initiate the closure. The rest we'll leave to the OS's TCP stack (and potentially other intermediates such as TLS wrappers) or the async runtime to figure out in the background.

Closes: #1309
Part of: DRIVERS-404

<details>

<summary>Reproducer</summary>

Assumes Ubuntu/Debian base.

0. Install dependencies `apt-get install iptables build-essential python3-venv python3-dev libnetfilter-queue-dev`
1. Start neo4j in docker, enable SSL for bolt, expose port 7687.
2. Create `filter.py`:
    ```python
    # requires
    # NetfilterQueue==1.1.0
    # scapy==2.7.0
    
    import signal
    import sys
    from netfilterqueue import NetfilterQueue
    from scapy.layers.inet import IP, TCP
    
    SERVER_IPS = {
        "127.0.0.1",
        "172.24.0.2",  # ADJUST: neo4j docker container IP
    }
    SERVER_PORT = 7687
    
    seen_syn_conns: set = set()  # client ports whose SYN was observed
    
    
    def conn_key(ip_pkt, tcp_pkt):
        if ip_pkt.dport == SERVER_PORT:
            return ip_pkt.sport
        return ip_pkt.dport
    
    
    def handle(pkt):
        raw = IP(pkt.get_payload())
        if not raw.haslayer(TCP):
            pkt.accept()
            return
    
        ip  = raw[IP]
        tcp = raw[TCP]
        key = conn_key(ip, tcp)
    
        if not (
            (ip.dst in SERVER_IPS or ip.src in SERVER_IPS)
            and (ip.dport == SERVER_PORT or ip.sport == SERVER_PORT)
        ):
            print(f"{raw} (unrelated)")
            pkt.accept()
            return
    
        print(f"{raw}")
    
        # Register the connection on the client's initial SYN.
        if "S" in tcp.flags:
            print(f"    [  syn] {key}")
            seen_syn_conns.add(key)
            pkt.accept()
            return
    
        # Drop any packet whose connection was never opened through us.
        if key not in seen_syn_conns:
            print(f"    [ drop] no SYN seen  {key}")
            pkt.drop()
            return
    
        pkt.accept()
    
    
    def shutdown(sig, _frame):
        print("\nShutting down.")
        sys.exit(0)
    
    
    signal.signal(signal.SIGINT,  shutdown)
    signal.signal(signal.SIGTERM, shutdown)
    
    nfq = NetfilterQueue()
    nfq.bind(1, handle)
    print("Running — Ctrl-C to stop.")
    try:
        nfq.run()
    finally:
        nfq.unbind()
    ```
3. Create venv for `filter.py` filter rule
   1. `python -m venv .venv`
   2. `source .venv/bin/activate`
   4. `pip install NetfilterQueue==1.1.0 scapy==2.7.0`
4. Start `sudo .venv/bin/python filter.py`
5. *Inside another* python environment with the driver and its dependencies available, run the following reproducer:
    ```python
    import sys
    import time
    import subprocess
    
    import asyncio
    from neo4j import AsyncGraphDatabase, GraphDatabase
    from neo4j.debug import watch
    
    watch("neo4j")
    
    URI = "bolt+ssc://localhost:7687"
    AUTH = ("neo4j", "pass")
    
    SPORT_FILTER = ("--sport", "7687")
    DPORT_FILTER = ("--dport", "7687")
    
    
    def _iptables_rule(port_filter):
        return (
            "INPUT", "-p", "tcp", *port_filter, "-j", "NFQUEUE", "--queue-num", "1"
        )
    
    
    def _clear_iptables():
        for port_filter in (SPORT_FILTER, DPORT_FILTER):
            subprocess.run(
                ["sudo", "iptables", "-D", *_iptables_rule(port_filter)],
                stderr=sys.stderr,
                stdout=sys.stdout,
            )
    
    
    def _setup_iptables():
        for port_filter in (SPORT_FILTER, DPORT_FILTER):
            subprocess.run(
                ["sudo", "iptables", "-I", *_iptables_rule(port_filter)],
                stderr=sys.stderr,
                stdout=sys.stdout,
                check=True,
            )
    
    
    async def main():
        _clear_iptables()
    
        async with AsyncGraphDatabase.driver(
            URI,
            auth=AUTH,
            max_connection_pool_size=1,
            connection_acquisition_timeout=3,
            liveness_check_timeout=1,
            max_connection_lifetime=5,
        ) as driver:
            t0 = time.time()
    
            try:
                async with driver.session() as session:
                    await (await session.run("RETURN 1 AS n")).to_eager_result()
            finally:
                print(f"\n  >>>>> took: {time.time() - t0:.2f} seconds\n")
    
            _setup_iptables()
            await asyncio.sleep(6)  # exceed connection lifetime
    
            t0 = time.time()
    
            try:
                async with driver.session() as session:
                    await (await session.run("RETURN 1 AS n")).to_eager_result()
            finally:
                print(f"\n  >>>>> took: {time.time() - t0:.2f} seconds\n")
    
    
    def main_sync():
        _clear_iptables()
    
        with GraphDatabase.driver(
            URI,
            auth=AUTH,
            max_connection_pool_size=1,
            connection_acquisition_timeout=3,
            liveness_check_timeout=1,
            max_connection_lifetime=5,
        ) as driver:
            t0 = time.time()
    
            try:
                with driver.session() as session:
                    session.run("RETURN 1 AS n").to_eager_result()
            finally:
                print(f"\n  >>>>> took: {time.time() - t0:.2f} seconds\n")
    
            _setup_iptables()
            time.sleep(1)  # exceed liveness check timeout
    
            t0 = time.time()
    
            try:
                with driver.session() as session:
                    session.run("RETURN 1 AS n").to_eager_result()
            finally:
                print(f"\n  >>>>> took: {time.time() - t0:.2f} seconds\n")
    
    
    if __name__ == "__main__":
        try:
            asyncio.run(main())
            # main_sync()
        finally:
            _clear_iptables()
            print(" ====== DONE ====== ")
    ```

</details>

### DIFF
--- a/src/neo4j/_async/io/_bolt.py
+++ b/src/neo4j/_async/io/_bolt.py
@@ -331,7 +331,7 @@ class AsyncBolt:
         except (ServiceUnavailable, SessionExpired, BoltHandshakeError):
             return None
         else:
-            await AsyncBoltSocket.close_socket(s)
+            AsyncBoltSocket.close_socket(s)
             return protocol_version
 
     @staticmethod
@@ -377,7 +377,7 @@ class AsyncBolt:
         bolt_cls = protocol_handlers.get(protocol_version)
         if bolt_cls is None:
             log.debug("[#%04X]  C: <CLOSE>", s.getsockname()[1])
-            await AsyncBoltSocket.close_socket(s)
+            AsyncBoltSocket.close_socket(s)
             raise UnsupportedServerProduct(
                 "The neo4j server does not support communication with this "
                 "driver. This driver has support for Bolt protocols "
--- a/src/neo4j/_async/io/_bolt.py
+++ b/src/neo4j/_async/io/_bolt.py
@@ -386,7 +386,7 @@ class AsyncBolt:
 
         try:
             auth = await AsyncUtil.callback(auth_manager.get_auth)
-        except Exception as e:
+        except (Exception, asyncio.CancelledError) as e:
             log.debug(
                 "[#%04X]  C: <CLOSE> open auth manager failed: %r",
                 s.getsockname()[1],
--- a/src/neo4j/_async/io/_bolt.py
+++ b/src/neo4j/_async/io/_bolt.py
@@ -386,21 +386,13 @@ class AsyncBolt:
 
         try:
             auth = await AsyncUtil.callback(auth_manager.get_auth)
-        except asyncio.CancelledError as e:
-            log.debug(
-                "[#%04X]  C: <KILL> open auth manager failed: %r",
-                s.getsockname()[1],
-                e,
-            )
-            s.kill()
-            raise
         except Exception as e:
             log.debug(
                 "[#%04X]  C: <CLOSE> open auth manager failed: %r",
                 s.getsockname()[1],
                 e,
             )
-            await s.close()
+            s.close()
             raise
 
         connection = bolt_cls(
@@ -998,17 +990,20 @@ class AsyncBolt:
             self.goodbye()
             try:
                 await self._send_all()
-            except (OSError, BoltError, DriverError) as exc:
+            except (
+                OSError,
+                BoltError,
+                DriverError,
+                SocketDeadlineExceededError,
+            ) as exc:
                 log.debug(
-                    "[#%04X]  _: <CONNECTION> ignoring failed close %r",
+                    "[#%04X]  _: <CONNECTION> ignoring failed final flush %r",
                     self.local_port,
                     exc,
                 )
         log.debug("[#%04X]  C: <CLOSE>", self.local_port)
         try:
-            await self.socket.close()
-        except OSError:
-            pass
+            self.socket.close()
         finally:
             self._closed = True
 
@@ -1019,13 +1014,7 @@ class AsyncBolt:
         log.debug("[#%04X]  C: <KILL>", self.local_port)
         self._closing = True
         try:
-            self.socket.kill()
-        except OSError as exc:
-            log.debug(
-                "[#%04X]  _: <CONNECTION> ignoring failed kill %r",
-                self.local_port,
-                exc,
-            )
+            self.socket.close()
         finally:
             self._closed = True
 
--- a/src/neo4j/_async/io/_bolt_socket.py
+++ b/src/neo4j/_async/io/_bolt_socket.py
@@ -350,7 +350,7 @@ class AsyncBoltSocket(AsyncBoltSocketBase):
                     err_str,
                 )
                 if s:
-                    await cls.close_socket(s)
+                    cls.close_socket(s)
                 errors.append(error)
                 failed_addresses.append(resolved_address)
             except asyncio.CancelledError:
@@ -366,7 +366,7 @@ class AsyncBoltSocket(AsyncBoltSocketBase):
                 raise
             except Exception:
                 if s:
-                    await cls.close_socket(s)
+                    cls.close_socket(s)
                 raise
         address_strs = tuple(map(str, failed_addresses))
         # TODO: 7.0 - when Python 3.11+ is the minimum, use exception groups
--- a/src/neo4j/_async/io/_bolt_socket.py
+++ b/src/neo4j/_async/io/_bolt_socket.py
@@ -177,7 +177,7 @@ class AsyncBoltSocket(AsyncBoltSocketBase):
             # If no data is returned after a successful select
             # response, the server has closed the connection
             log.debug("[#%04X]  S: <CLOSE>", ctx.local_port)
-            await self.close()
+            self.close()
             raise ServiceUnavailable(
                 f"Connection to {ctx.resolved_address} closed with incomplete "
                 f"handshake response"
@@ -185,7 +185,7 @@ class AsyncBoltSocket(AsyncBoltSocketBase):
         if data_size != n:
             # Some garbled data has been received
             log.debug("[#%04X]  S: @*#!", ctx.local_port)
-            await self.close()
+            self.close()
             raise BoltProtocolError(
                 f"Expected {ctx.ctx} from {ctx.resolved_address!r}, received "
                 f"{response!r} instead (so far {ctx.full_response!r}); "
@@ -264,7 +264,7 @@ class AsyncBoltSocket(AsyncBoltSocketBase):
 
         if response == b"HTTP":
             log.debug("[#%04X]  C: <CLOSE> (received b'HTTP')", local_port)
-            await self.close()
+            self.close()
             raise ServiceUnavailable(
                 f"Cannot to connect to Bolt service on {resolved_address!r} "
                 "(looks like HTTP)"
@@ -362,8 +362,7 @@ class AsyncBoltSocket(AsyncBoltSocketBase):
                     "[#%04X]  C: <CANCELED> %s", local_port, resolved_address
                 )
                 if s:
-                    with suppress(OSError):
-                        s.kill()
+                    s.close()
                 raise
             except Exception:
                 if s:
--- a/src/neo4j/_async_compat/network/_bolt_socket.py
+++ b/src/neo4j/_async_compat/network/_bolt_socket.py
@@ -227,7 +227,7 @@ class AsyncBoltSocketBase(abc.ABC):
         self._writer.write(data)
         return await self._wait_for_write(self._writer.drain)
 
-    def close(self):
+    def close(self) -> None:
         # Simulate `SO_LINGER` off:
         # flush data and close socket in the background, don't block
         with suppress(OSError):
@@ -383,14 +383,14 @@ class AsyncBoltSocketBase(abc.ABC):
     ) -> tuple[t.Self, BoltProtocolVersion]: ...
 
     @classmethod
-    async def close_socket(cls, socket_):
+    def close_socket(cls, socket_: t.Self | socket) -> None:
         if isinstance(socket_, AsyncBoltSocketBase):
             socket_.close()
         else:
             cls._kill_raw_socket(socket_)
 
     @classmethod
-    def _kill_raw_socket(cls, socket_):
+    def _kill_raw_socket(cls, socket_: socket) -> None:
         with suppress(OSError):
             socket_.shutdown(SHUT_RDWR)
         with suppress(OSError):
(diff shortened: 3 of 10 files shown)


MaxAake (approved):
✅

Issue #1310, opened 2026-05-29
Unguarded getsockname()[1] on async connection open → unretryable TypeError: 'NoneType' object is not subscriptable (deterministic repro; re: #730, #1057)


riley-deliberately:
## AI DISCLOSURE 
I used Claude Code to generate this repro.  Its very hard to verify that the socketname is really "None" in the running system, but by patching this code in my production deployment the error did go away...

## The Issue

`AsyncBolt.__init__` (and several sibling sites in the async connect path) call `self.socket.getsockname()[1]` with no `None` guard. For the async driver, `BoltSocket.getsockname()` is `self._writer.transport.get_extra_info("sockname")`, which asyncio returns as **`None`** once the transport's socket is gone — i.e. when a load balancer drops a freshly-opened connection (we see this constantly on **Neo4j Aura**, most often on the connection opened for a **routing-table refresh**).

Because the result is a bare `TypeError` rather than `ServiceUnavailable`/`SessionExpired`, the driver's transaction-retry never catches it, so a *transient* connection drop becomes a hard, non-retryable crash with a misleading message.

This is the same defect previously reported in **#730** (2022) and **#1057** (2024); both were closed because the condition **could not be reproduced** (#1057: *"I was not able to reproduce this condition locally"*, and *"feel free to reopen … while providing the additional information requested"*). **This report provides a deterministic, minimal reproduction** — the missing piece — plus a verification that guarding the `None` fixes it.

### Versions

- Driver: reproduced on `neo4j==5.28.4`; the unguarded line is **identical in the latest `6.2.0`** (`_bolt.py:159`).
- Server: Neo4j Aura in production; the repro below works against any local Neo4j.
- Python 3.12.

### Production traceback (matches #730 / #1057)

```
File ".../neo4j/_async/io/_pool.py", line 792, in fetch_routing_info
    cx = await self._acquire(address, auth, deadline, None)
File ".../neo4j/_async/io/_pool.py", line 711, in opener
    return await AsyncBolt.open(...)
File ".../neo4j/_async/io/_bolt.py", line 413, in open
    connection = bolt_cls(...)
File ".../neo4j/_async/io/_bolt.py", line 156, in __init__
    self.local_port = self.socket.getsockname()[1]
                      ~~~~~~~~~~~~~~~~~~~~~~~~~^^^
TypeError: 'NoneType' object is not subscriptable
```

### Deterministic reproduction

[...]

PR #1313, opened 2026-06-05
[5.28] Cache sockname and peername

robsdedude:
Async
-----
Instead of relying on asyncio's caching of these properties, we do it ourselves. The advantage is that asyncio populates the cache on a best-effort basis. This can lead to the values being `None` if retrieving them causes an `OSError`.

This is a misalignment between the async and sync driver that this PR aims to remedy.

Sync
----
While in the async driver we're introducing (custom) caching where implicit caching was already in place, we also introduce caching of these fields in the sync driver for parity (and potentially fewer syscalls).

Fixes: #1310
Part of: DRIVERS-418

### DIFF
--- a/src/neo4j/_async_compat/network/_bolt_socket.py
+++ b/src/neo4j/_async_compat/network/_bolt_socket.py
@@ -100,7 +100,7 @@ def _non_expired_timeout(
 class AsyncBoltSocketBase(abc.ABC):
     Bolt: te.Final[type[AsyncBolt]] = None  # type: ignore[assignment]
 
-    def __init__(self, reader, protocol, writer) -> None:
+    def __init__(self, reader, protocol, writer, sockname, peername) -> None:
         self._reader = reader  # type: asyncio.StreamReader
         self._protocol = protocol  # type: asyncio.StreamReaderProtocol
         self._writer = writer  # type: asyncio.StreamWriter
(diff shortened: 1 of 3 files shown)


src/neo4j/_async_compat/network/_bolt_socket.py
```diff
@@ -374,21 +380,24 @@ def _socket(self) -> socket | SSLSocket:
     @_socket.setter
     def _socket(self, socket_: socket | SSLSocket) -> None:
         self.__socket = socket_
-        self.getsockname = socket_.getsockname
-        self.getpeername = socket_.getpeername
         if hasattr(socket, "getpeercert"):
             self.getpeercert = t.cast(SSLSocket, socket_).getpeercert
         elif "getpeercert" in self.__dict__:
             del self.__dict__["getpeercert"]
+            socket_.getsockname()
```
AndyHeap-NeoTech: Is there a reason that this is called here? The return value is not used. Possible leftover stray line?

AndyHeap-NeoTech (commented):
Looks good, just one small comment for clarification.

src/neo4j/_async_compat/network/_bolt_socket.py
robsdedude: Good catch :bow:

AndyHeap-NeoTech (approved):
LGTM.

robsdedude:
> [!IMPORTANT]
> **NOT READY FOR MERGE**
> Waiting for user feedback to confirm/disprove proposed fixed.

PR #1314, opened 2026-06-05
[5.28] Make mypy happy again

robsdedude:
This method is public-ish, i.e., reachable without accessing any `_xyz`, but it's undocumented. The driver stopped using it a while ago. This change is just to please the type-checker.


StephenCathcart (approved):
👍

PR #1316, opened 2026-06-08
Fix pyarrow import

robsdedude:
`neo4j.vector.Vector` uses `pyarrow.compute.count` but the driver never loads the module `pyarrow.compute`. We've not seen issues in tests because we either have all optional dependencies or none installed. And the current versions of `pandas` (another optional dependency), when loaded, in turn loads `pyarrow.compute`. However, this is not guaranteed to remain this way and won't work for users only that only have pyarrow (but not pandas) installed.

Further does this PR silence pyarrow's type annotations for now as they're not quite up to snuff yet.

Closes: DRIVERS-443

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -225,6 +225,13 @@ module = [
 ]
 ignore_missing_imports = true
 
+[[tool.mypy.overrides]]
+# https://github.com/apache/arrow/issues/50123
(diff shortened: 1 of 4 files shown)


StephenCathcart (approved):
👍

PR #1317, opened 2026-06-23
Tests: fix deprecations

robsdedude:
* `pytest` deprecated using generators (lazy) for test parametrization. Instead a collection (eager) such as tuple, or list is expected.
* `numpy` deprecated instantiating temporal type values without providing an explicit unit.


MaxAake (approved):
✅

PR #1319, opened 2026-07-07
HTTP Query API Support

robsdedude:
Neo4j servers recently introduced a new HTTP API for querying the database (called Query API): https://neo4j.com/docs/query-api/current/

This PR add support for configuring the driver with a HTTP URL (e.g. `https://neo4j.example.com:7474/`), which will make the driver run queries and transactions via the HTTP Query API instead of Bolt.

### Notes
Please note that:
 * This feature is in preview (see below).
 * For the HTTP support 3rd party libraries (currently `urllib3` for sync and `aiohttp` for async) are required. Install `pip install neo4j[http]` instead of `pip install neo4j`.
 * Some features are not supported and likely won't be in the future:
   * home database resolution (i.e., user code must always provide an explicit database name for each session created)
   * the only supported authentication scheme is `basic`
 * Some features are not yet supported:
   * Transaction metadata and timeouts are being ignored.
   * Some details in the summary (such as `neo4j.ResultSummary.server.protocol_info`) might be missing or inaccurate.
   * Vector types, the `UnsupportedType` type, and `UUID`s cannot be exchanged with the DBMS.
   * Notifications and errors have fallback GQL status codes and status information.
   * `fetch_size` has not effect, there is no backpressure mechanism.
   * Notification filters (`notifications_min_severity`, `notifications_disabled_categories`, `notifications_disabled_classifications`) have no effect.

### Preview
This feature is in [*preview*](https://github.com/neo4j/neo4j-python-driver/wiki/preview-features). This means that it does not follow semver. It might be changed (or in rare cases be removed) at any time without following the normal deprecation cycle.

---

Depends on:
 * https://github.com/neo4j-drivers/testkit/pull/720

Closes: DRIVERS-118


testkit/testkit.json
```diff
@@ -1,6 +1,6 @@
 {
   "testkit": {
     "uri": "https://github.com/neo4j-drivers/testkit.git",
-    "ref": "6.x"
```
robsdedude, line 4: TODO: revert once TestKit PR has been merged:

 * https://github.com/neo4j-drivers/testkit/pull/720

src/neo4j/_async/driver.py
```diff
@@ -481,6 +502,10 @@ def _parse_targets(cls, *targets):
         )
 
 
+class _Http(_Direct):
+    _default_port = 7687
```
StephenCathcart: `_Http` uses `_default_port = 7687`, the bolt port, so `http://localhost` with no port goes there rather than on 7474:

    driver = neo4j.GraphDatabase.driver("http://localhost", auth=AUTH)
    driver.address.port  # -> 7687

is that intended, or does https want its own default (7473) as well?

src/neo4j/_async/io/_http/_http2.py
```diff
@@ -0,0 +1,1390 @@
...
+                    "[#%04X]  _: Connection error: %r",
+                    self._id,
+                    error,
+                )
+                self._defunct = True
+                if user_cancelled:
+                    raise
+                if protocol_error or not connection_failed:
+                    raise
+                for request in self._requests:
```
StephenCathcart: `_CommitRequest` never looks constructed anywhere (i think), so this isinstance can't match. the loop also scans the queued `_requests`, and `send_all()` has already popped the one in flight (the bolt version looks at `self.responses` for `CommitResponse` instead i believe).

i pointed the driver at a stub that drops the connection on the commit POST and answers the retry, and the write went in twice:

    write query submitted : 2 time(s)
    commit attempts       : 2
    commits dropped       : 1
    execute_write returned ['Alice']

`ServiceUnavailable.is_retryable()` is True where `IncompleteCommit`'s is False, so the transaction function replays a commit with an outcome we don't know. am i reading that right? happy to hand over the stub i used to test it

src/neo4j/_async_compat/concurrency.py
```diff
@@ -475,8 +477,24 @@ def notify_all(self):
...
+
 Condition: t.TypeAlias = threading.Condition
 CooperativeLock: t.TypeAlias = threading.Lock
 Lock: t.TypeAlias = threading.Lock
 CooperativeRLock: t.TypeAlias = threading.RLock
 RLock: t.TypeAlias = threading.RLock
+BoundedSemaphore: t.TypeAlias = threading.BoundedSemaphore
+
+
+def acquire_bounded_semaphore(
```
StephenCathcart, line 521: `semaphore.acquire(timeout=...)` returns False on timeout rather than raising, and the result isn't kept here, so with `max_connection_pool_size=1` i get a second connection out of the sync pool:

    conn 0: acquired
    conn 1: acquired          <- over the limit
    close:  ValueError: Semaphore released too many times

the async twin does block, but comes out as a bare `TimeoutError` rather than `ConnectionAcquisitionTimeoutError` like the bolt pool. should this be checking the return value, or is this a none issue?

src/neo4j/_async/io/_http/_base.py
```diff
@@ -0,0 +1,228 @@
...
+    _MAX = 2**16
+
+    def __init__(self) -> None:
+        self._next_id = 1
+        self._lock = AsyncCooperativeLock()
+
+    async def next_id(self) -> int:
+        async with self._lock:
+            current_id = self._next_id
+            self._next_id = min((self._next_id + 1) % IdGenerator._MAX, 1)
```
StephenCathcart: `min` keeps this at 1 forever, so every http connection logged for me as `[#0001]`. did you mean `max`, to skip 0 on wrap?

```suggestion
            self._next_id = max((self._next_id + 1) % IdGenerator._MAX, 1)
```

src/neo4j/_io.py
```diff
@@ -98,6 +106,51 @@ def __str__(self) -> str:
         return f"{self.major}.{self.minor}"
 
 
+_VERSION_RE = re.compile(r"^(\d+)\.(\d+)")
+
+
+@dataclasses.dataclass(slots=True, frozen=True, kw_only=True)
+class HTTPServerInfo:
+    neo4j_version: str
+    parsed_neo4j_version: tuple[int, int] | None = dataclasses.field(
```
StephenCathcart, line 115: i've got a half-memory of a call about parsing the server version and rejecting the older ones that send `values` as a flat list rather than a list of records, so users get a clear message instead of a parse error. `parsed_neo4j_version` is computed here but i couldn't find anything reading it, so i may have the wrong end of this.

a stub advertising 5.19.0 with the flat shape gives:

    QueryApiHttpError: protocol error: expected list, got: 1

and that class's docstring points people at filing a driver bug. can't recall if I'm remembering this correctly, and whether it still needs changed?

docs/source/async_api.rst
```diff
@@ -10,8 +10,8 @@ Async API Documentation
 AsyncGraphDatabase
 ******************
 
-Async Driver Construction
-=========================
+AsyncDriver Construction
+========================
```
StephenCathcart, line 14: `async_api.rst` doesn't look like it picked up the http changes that `api.rst` got: no `http`/`https` in the valid-URI list or the scheme table, no `AsyncHttpDriver` autoclass section, and line 454 still reads "the URI scheme is ``bolt://`` or ``neo4j://``". `AsyncHttpDriver` is exported and carries the whole "Unavailable Features" docstring, so i think none of it renders for async users. are the async docs to do at a later time?

src/neo4j/_async/driver.py
```diff
@@ -1373,6 +1398,117 @@ def __init__(self, pool, default_workspace_config):
...
+        * :attr:`.ServerInfo.agent` is being computed from the DBMS's
+          advertised version. Further, it is being cached to reduce
+          round-trips and overloading the DBMS's HTTP endpoints.
+        * :attr:`.ResultSummary.query_type` will always be :data:`None`.
+
+    * Transmitting and receiving :class:`Vector` values is currently not
+      supported.
+
+    * The only supported auth scheme (see :ref:`auth-ref`) is ``"basic"`` and
+      ``"bearer"``.
```
StephenCathcart, line 1475:
```suggestion
        * :attr:`.ResultSummary.result_available_after` and
          :attr:`.ResultSummary.result_consumed_after` will always be
          :data:`None` because this information is not provided by the server.
        * :attr:`.ServerInfo.agent` is being computed from the DBMS's
          advertised version. Further, it is being cached to reduce
          round-trips and overloading the DBMS's HTTP endpoints.
        * :attr:`.ResultSummary.query_type` will always be :data:`None`.

    * Transmitting and receiving :class:`Vector` values is currently not
      supported.

    * The only supported auth schemes (see :ref:`auth-ref`) are ``"basic"`` and
      ``"bearer"``.
```

src/neo4j/_async/io/_connection.py
```diff
@@ -0,0 +1,375 @@
...
+    @property
+    @abc.abstractmethod
+    def supports_re_auth(self) -> bool:
+        """Whether the connection version supports re-authentication."""
+        raise NotImplementedError
+
+    @property
+    @abc.abstractmethod
+    def supports_notification_filtering(self) -> bool:
+        """Whether the connection version supports re-authentication."""
```
StephenCathcart:
```suggestion
        """Whether the connection version supports notification filtering."""
```

docs/source/api.rst
```diff
@@ -1101,8 +1161,11 @@ Specifically, the following applies:
 
     This will use the default database on the Neo4j instance.
 
-:Type: ``str``, ``neo4j.DEFAULT_DATABASE``
+.. note::
+    A string value must be provided connected via
+    ``http://`` or ``https://`` scheme.
```
StephenCathcart:
```suggestion
    A string value must be provided when connected via
    ``http://`` or ``https://`` scheme.
```

src/neo4j/_async/io/_http/_base.py
```diff
@@ -0,0 +1,228 @@
...
+            query_api = await factory.new_http_query_api(
+                self._address,
+                self._path,
+                pool_config=pool_config,
+            )
+            try:
+                return await self._fetch_unguarded(query_api)
+            except Exception as e:
+                log.warning(
+                    "[#%04X]  _: Failed to fetch  server info: %r",
```
StephenCathcart:
```suggestion
                    "[#%04X]  _: Failed to fetch server info: %r",
```

src/neo4j/_async_compat/network/_http_query_api.py
```diff
@@ -0,0 +1,567 @@
...
+            self._repr = self._compute_repr()
+        return self._repr
+
+    def _compute_repr(self) -> str:
+        fields_repr = (
+            f"{key!r}: {value!r}" for key, value in self._iter_sanintized()
+        )
+        return f"{{{', '.join(fields_repr)}}}"
+
+    def _iter_sanintized(self) -> t.Generator[tuple[str, str]]:
```
StephenCathcart:
```suggestion
    def _compute_repr(self) -> str:
        fields_repr = (
            f"{key!r}: {value!r}" for key, value in self._iter_sanitized()
        )
        return f"{{{', '.join(fields_repr)}}}"

    def _iter_sanitized(self) -> t.Generator[tuple[str, str]]:
```

docs/source/api.rst
```diff
@@ -906,6 +952,19 @@ Will result in:
 .. autoclass:: neo4j.Neo4jDriver
 
 
+.. _http-driver-ref:
+
+HttpDriver
+===========
```
StephenCathcart:
```suggestion
==========
```

src/neo4j/_async/driver.py
StephenCathcart: `is` -> `are`, and removed the TODO as I "think" it's `None` - worth checking

docs/source/api.rst
StephenCathcart: ocd

StephenCathcart (approved):
Other than open questions looks good to me 👍

src/neo4j/_async/io/_http/_http2.py
StephenCathcart: raised this on testkit#720 too, with a test that catches it, so no need for the stub now

src/neo4j/_io.py
robsdedude: I remember we discussed something along those lines 
 1. during refinement 2026-04-28 we decided to fail hard if the discovery endpoint does not exist or if the server version can't be parsed.
 2. However, I don't think we made a call on which server versions we wouldn't want to support/reject (only that we'd draw the line somewhere™ TBD).
 3. Neither do I remember that we made a decision on whether older servers should fail hard (i.e., driver perform the check you propose here) or fail soft (i.e., the moment they misbehave, anything may happen including protocol violation errors).

Can you find notes or similar to remind me if anything regarding 2. or 3. was decided?

PR #1330, opened 2026-08-26
[5.28] Help mypy with explicit type annotation

robsdedude:
Partial back port of: https://github.com/neo4j/neo4j-python-driver/pull/1327

Part of: DRIVERS-526


robsdedude:
Fixes the type checking only. Failing tests will be tackled separately in https://github.com/neo4j/neo4j-python-driver/pull/1331.

injectives (approved):

PR #1335, opened 2026-09-01
Clean up dead and broken code

robsdedude:
The code is broken because
 * in async, `AsyncBolt.socket` is of type `AsyncBoltSocket` wrapping `asyncio` constructs
 * in sync, `Bolt.socket` is of type `BoltSocket` wrapping `socket.socket` or `ssl.SSLSocket`

Note that `(Async)Bolt.socket` is *never* of type `ssl.SSLSocket` directly. Therefore this check is useless. The code could be fixed, but it's never called. Therefore removing it is the best option.

### DIFF
--- a/src/neo4j/_async/io/_bolt.py
+++ b/src/neo4j/_async/io/_bolt.py
@@ -432,11 +432,6 @@ class AsyncBolt:
 
     @property
     @abc.abstractmethod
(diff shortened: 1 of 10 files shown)


MaxAake (approved):
✅

PR #1338, opened 2026-09-02
[5.28] Benchkit: fix type-guard checking wrong variable

robsdedude:
Back port of: https://github.com/neo4j/neo4j-python-driver/pull/1336


MaxAake (approved):
✅

PR #1334, opened 2026-09-01
Fix BoltSocketBase.getpeercert pass-through

robsdedude:
`(Async)BoltSocketBase.getpeercert` is not used throughout the code-base. Therefore does this bug and its fix have no direct impact. It's future-proofing the code-base.

The `hasattr` was introspecting the `socket.socket` class instead of the `socket_` object. A silly typo, really.

PR #1339, opened 2026-09-02
[5.28] Fix BoltSocketBase.getpeercert pass-through

robsdedude:
`(Async)BoltSocketBase.getpeercert` is not used throughout the code-base. Therefore does this bug and its fix have no direct impact. It's future-proofing the code-base.

The `hasattr` was introspecting the `socket.socket` class instead of the `socket_` object. A silly typo, really.

(Boiled-down) back port of: https://github.com/neo4j/neo4j-python-driver/pull/1334


MaxAake (approved):
✅

PR #1344, opened 2026-09-14
Ruff config: revert to rule codes for now

robsdedude:
Makes ruff show rule codes (like `RUF050`) instead of rule names (like `unnecessary-if`). This is consistent with us using codes to (de)select rules and mute them in comments.

Amends: https://github.com/neo4j/neo4j-python-driver/pull/1342


MaxAake (approved):
✅

PR #1346, opened 2026-09-23
Fix race condition in bolt connection pool

robsdedude:
The pool tries to acquire an idle connection first. If no idle connection is
available, it puts itself into a queue waiting for another thread/task to return
a connection before trying again.

However, there is no protection against a connection being returned between the
attempt to acquire an idle connection and the enqueuing. Therefore, it could
happen that a thread waits indefinitely like this:

Assume max pool size of 1

 * T1 creates & acquires con1
 * T2 checks for idle connections => None
 * T1 releases con1
 * T2 acquires lock to be able to wait for the release event
 * T1 terminates
 * T2 times out waiting for Godot
 
 Closes: DRIVERS-578


StephenCathcart (approved):
👍 looks good

